Issue #14, opened 2020-07-10
Bug: Wrong commit count


anshumanv:
Hey 👋 

Cool project, is there a limitation to the number of commits/PRs it can fetch?

They seem to be off by quite a lot for instance - 

![image](https://user-images.githubusercontent.com/21009455/87129089-6acae000-c2ae-11ea-8edc-b46abce403a2.png)

In normal GitHub search - 
![image](https://user-images.githubusercontent.com/21009455/87129136-7d451980-c2ae-11ea-8033-2674ad204b62.png)

anuraghazra:
Hmm it should be fine.. it fetches the data from Github GraphQL api so it can't be invalid. although it is possible that you contributed to github organizations and i think those are not counted. #1 

![Stat](https://github-readme-stats.vercel.app/api?username=anshumanv)

anshumanv:
Ah okay, very strange because PRs count is close to correct but commit count is still far off even if you exclude org commits.

anuraghazra:
@anshumanv i think i got it. the issue is that github only returns last years contribution commits 

```gql

{
  user(login: "anshumanv") {
   contributionsCollection(from: "2016-08-13T14:31:00Z", to:"2020-07-10T08:35:06.094Z") {
      totalCommitContributions
    }
  }
}

```


I can set from and to variables in the query but github does not allow more than one year's contributions to be shown at once

I have to find a way to fix this. do you have any ideas on this?

anuraghazra:
@anshumanv what i did is that looped through all the years and fetched the commits individually and seems kinda close. 

![github-readme-issue](https://user-images.githubusercontent.com/35374649/87137317-51c82c00-c2ba-11ea-926d-29daf764c8dc.PNG)

anshumanv:
Yes, this is the basic brute force solution but I think you'll run into API rate limit after a while if you do this for every user.

anshumanv:
Maybe you can give an option for users to use their own tokens?

anuraghazra:
No i don't think i can hit the rate limit because it's already authenticated by the server... i auth token in env.

anuraghazra:
Also i don't think i can give users an option to add their own tokens because that would be a huge security vulnrability if someone did like this `/api?username=abc&token=SECRET_TOKEN` because the url would be public

anshumanv:
> No i don't think i can hit the rate limit because it's already authenticated by the server... i set auth token in env.

[...]

PR #15, opened 2020-07-10
fix: total commits count

anuraghazra:
So as i mentioned in #14 that github api only return commits of 1 year max. what i did is that fetched all user years and then looped through that and made a graphql request for each year 

```gql
    data: {
          query: `
          query userInfo($login: String!, $from: DateTime!, $to: DateTime!) {
            user(login: $login) {
              contributionsCollection(from: $from, to: $to) {
                totalCommitContributions
              }
            }
          }
        `,
          variables: {
            login: username,
            from: currentDate.toISOString(), // current year
            to: nextDate.toISOString(), // next year
          },
        },
```

TODO: perf improvement?

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -56,11 +56,17 @@ async function fetchStats(username) {
   // FETCH ALL COMMITS
   let commitPromises = user.contributionsCollection.contributionYears.map(
     async (year) => {
-      let currentDate = new Date();
+      // Don't fetch contributions older than 2008 (Perf optimization)
+      if (year < 2008) {
+        return {
+          user: { contributionsCollection: { totalCommitContributions: 0 } },
+        };
+      }
+
+      const currentDate = new Date();
       currentDate.setFullYear(year, 0, 0);
-      let nextDate = new Date();
-      nextDate.setFullYear(year + 1, 0, 0);
-      let res = await axios({
+
+      const res = await axios({
         url: "https://api.github.com/graphql",
         method: "post",
         headers: {
@@ -68,9 +74,9 @@ async function fetchStats(username) {
         },
         data: {
           query: `
-          query userInfo($login: String!, $from: DateTime!, $to: DateTime!) {
+          query userInfo($login: String!, $from: DateTime!) {
             user(login: $login) {
-              contributionsCollection(from: $from, to: $to) {
+              contributionsCollection(from: $from) {
                 totalCommitContributions
               }
             }
@@ -79,10 +85,10 @@ async function fetchStats(username) {
           variables: {
             login: username,
             from: currentDate.toISOString(),
-            to: nextDate.toISOString(),
           },
         },
       });
+
       console.log(
         currentDate.toDateString(),
         res.data.data.user.contributionsCollection.totalCommitContributions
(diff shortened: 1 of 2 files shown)


api/index.js
```diff
@@ -53,9 +53,61 @@ async function fetchStats(username) {
 
   const user = res.data.data.user;
 
+  // FETCH ALL COMMITS
+  let commitPromises = user.contributionsCollection.contributionYears.map(
+    async (year) => {
+      // Don't fetch contributions older than 2008 (Perf optimization)
```
anshumanv, line 59: ah thanks 💯 😆

anuraghazra: Yup! fixed it real quick. since github was first started on 2008 i wouldn't care about older commits.

api/index.js
```diff
@@ -53,9 +53,61 @@ async function fetchStats(username) {
...
+          }
+        `,
+          variables: {
+            login: username,
+            from: currentDate.toISOString(),
+          },
+        },
+      });
+
+      console.log(
```
anshumanv, line 92: rm log?

anuraghazra: Hmmm yeah probably i'll remove... that although it's fine too helps me debug the serverless function with vercel logs

anuraghazra:
@anshumanv Hi, i just did some performance testing and found out that to fetch all of your profile's commits it took over ~4 seconds to get back the response.. 

So do you have any ideas of improving the speed? 

One thing i can do is have a option for specifying if users want to fetch all of the commits like `?fetch_all_commits=true`

anshumanv:
> One thing i can do is have a option for specifying if users want to fetch all of the commits like ?fetch_all_commits=true

This is definitely one but I'm not sure of cases when a user would want to limit their count, but can't think of anything else atm. This seems the best choice to me. 👍

api/index.js
anshumanv: Then you should probably only do it in case of exceptions which need debuggin, else you're just logging normal queries unless that's what you're intending

anshumanv:
You can give it a count param too, for the number of queries i.e no of years.

kingthorin:
Why not change the default text to "Current Year Commits" or "Commits for Last 365d" or something. With an option for "Total" commits along with a statement that there's a performance concern in doing such....

TheAshwanik:
is this fixed and merged to master branch?

anuraghazra:
No @TheAshwanik meanwhile you can hide the totalCommits stats if you want with `hide=["commits"]`

TheAshwanik:
> commits

Thanks.. My problem is i like the stats and I want to improve my rank but, if the commit count is not correct then the rank also would not be correct right?

anuraghazra:
Don't worry about rank it's not something that shows how skilled or how good you are. its just an rank for github activity. just like contribution chart. also fix would come up in this week, we need to fix the rank algo too because its biased currently

TheAshwanik:
> Don't worry about rank it's not something that shows how skilled or how good you are. its just a rank for github activity. just like contribution chart. also fix would come up in this week, we need to fix the rank algo too because its biased currently

Yes. Absolutely.
Thanks for your brilliant work.

anuraghazra:
Fixed in #211

PR #19, opened 2020-07-10
Use destructuring query assignment

omidnikrah:


anuraghazra:
@omidnikrah Thanks for the improvement!

PR #22, opened 2020-07-11
Add cache to function response

nombrekeff:
I tried adding a cache (1h, but can be changed) to help with the commit count problem. It reduced Anshuman Verma's request from 4.5s avg to 185s avg in subsequent requests. The first request took a bit longer. 
This approach has a problem, though. The stats will only refresh once the cache max-age has expired, I guess it's not that big of a deal, shield.io and similar tend to have cache in place.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -189,7 +189,7 @@ module.exports = async (req, res) => {
 
   const stats = await fetchStats(username);
 
-  res.setHeader('Cache-Control', 's-maxage=300');
+  res.setHeader('Cache-Control', 'public, max-age=300');
   res.setHeader("Content-Type", "image/svg+xml");
   res.send(
     renderSVG(stats, {
--- a/api/index.js
+++ b/api/index.js
@@ -189,7 +189,7 @@ module.exports = async (req, res) => {
 
   const stats = await fetchStats(username);
 
-  res.setHeader('Cache-Control', 's-maxage=3600');
+  res.setHeader('Cache-Control', 's-maxage=300');
   res.setHeader("Content-Type", "image/svg+xml");
   res.send(
     renderSVG(stats, {
--- a/api/index.js
+++ b/api/index.js
@@ -189,6 +189,7 @@ module.exports = async (req, res) => {
 
   const stats = await fetchStats(username);
 
+  res.setHeader('Cache-Control', 's-maxage=3600');
   res.setHeader("Content-Type", "image/svg+xml");
   res.send(
     renderSVG(stats, {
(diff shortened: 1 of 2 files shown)


fransallen:
This one is legit!

Maybe consider using `public, max-age=` instead of `s-maxage`-only for general cache support.

api/index.js
```diff
@@ -189,6 +189,7 @@ module.exports = async (req, res) => {
 
   const stats = await fetchStats(username);
 
+  res.setHeader('Cache-Control', 's-maxage=3600');
```
anuraghazra, line 192: I think 5min cache would be good enough since we need the frequent updates too

res.setHeader("Cache-Control", "public, max-age=300");

fransallen: Yep, [similar](https://github.com/fransallen/github-readme-stats/blob/master/api/index.js#L21)

nombrekeff: 5 minutes would be good I guess, I made it one hour just to test it.

nombrekeff:
I lowered it to 5 minutes. Is there any other place it should be added? I guess it's only needed because of the commit count, but might be worth adding to the pin request. What you think?

nombrekeff:
> This one is legit!
> 
> Maybe consider using `public, max-age=` instead of `s-maxage`-only for general cache support.

What is the difference between the 2? I have not read much about it.

I can change it to `public, max-age=` if necessary

anuraghazra:
@nombrekeff I think you can also add cache headers to pin.js file. 

Also let's use public, max-age since I checked it works. I did not checked the other one

nombrekeff:
Sure, I've been reading about it, and `public, max-age` seems a better approach. I will change it and add it to pin.js as well.

nombrekeff:
Changed and added it to `pin.js` as well

api/index.js
dikshantpatodia: Ah! magic number

anuraghazra:
@nombrekeff LGTM! thanks for the PR :heart:

anuraghazra:
Opps looks like the PR got merged with the commit-counts branch, no issues, i think its good i can work on that branch and merge it directly.

nombrekeff:
Great! Hope it helps 😄

api/index.js
nombrekeff: yup, good catch, I can move it into a constant or something.

```js
const FIVE_MINUTES = 300;
res.setHeader('Cache-Control', `public, max-age=${FIVE_MINUTES}`);
```

dikshantpatodia: 1. Create a constants file under src, add all the constants there -> export -> tada

nombrekeff: yup, I commented that to Anurag, but he said he wanted to add that to env so it can easily be changed.

dikshantpatodia: Awesome

PR #29, opened 2020-07-11
add request utils + remove duplicate codes

omidnikrah:


anuraghazra:
Thanks for the PR @omidnikrah

src/fetchRepo.js
```diff
@@ -1,49 +1,39 @@
...
-            id
-            name
-          }
-          forkCount
+  const res = await request(`
+    fragment RepoInfo on Repository {
+      name
+      stargazers {
+        totalCount
+      }
```
anuraghazra, line 13: Can we change the API to be like this?

```js
const res = await request({
  query: `...`,
  variables: {...}
})
```

because it's more explicit and i like explicitness

omidnikrah: Yes, sure!

anuraghazra:
LGTM! thanks @omidnikrah

PR #45, opened 2020-07-13
fix: firefox layout breaking

anuraghazra:
fixes #45

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -1,5 +1,4 @@
 require("dotenv").config();
-const useragent = require("useragent");
 const { renderError } = require("../src/utils");
 const fetchStats = require("../src/fetchStats");
 const renderStatsCard = require("../src/renderStatsCard");
--- a/api/index.js
+++ b/api/index.js
@@ -19,9 +19,6 @@ module.exports = async (req, res) => {
   } = req.query;
   let stats;
 
-  const uaHeader = req.headers && req.headers["user-agent"];
-  const isFirefox = useragent.is(uaHeader).firefox;
-
   res.setHeader("Content-Type", "image/svg+xml");
   try {
     stats = await fetchStats(username);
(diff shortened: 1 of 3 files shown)


anuraghazra:
@saurabhdaware can you check the updates?

here's live development preview

cjdenio:
It's definitely better, but still a bit off.

image

anuraghazra:
@cjdenio Yes i checked it, i have no idea why tho, its working fine locally. ff has lot of issues with SVGs :weary: weird

cjdenio:
Weird, it works perfectly when icons are turned on.
image

anuraghazra:
Huh! very weird

anuraghazra:
Oh got it, for some reason user-agent isn't recognized

saurabhdaware:
Yes I have same output as @cjdenio

anuraghazra:
Ahh got it, the problem is that in the request headers i'm getting `github-camo` instead of `firefox's ua` which makes sense because the API is hit from server not directly from browser, any ideas to fix this? @saurabhdaware @cjdenio

saurabhdaware:
Can we fix that on frontend? https://stackoverflow.com/questions/5378559/including-javascript-in-svg

anuraghazra:
Not that easy i assume, the SVGs wont gonna be rendered as SVG or pure HTML elements by github, github will automatically upload the svg to their camo.githubusercontent cdn and then render that as image.

saurabhdaware:
ahh yes right

anuraghazra:
@saurabhdaware can you check again it should be fixed :D

cjdenio:
image

Looks perfect!

anuraghazra:
Awesome! thanks @cjdenio

saurabhdaware:
uh oh I turned off my laptop and sleeping now :( so probably by tomorrow morning

anuraghazra:
No issues @saurabhdaware @cjdenio sent the screenshot its working fine now, thanks for helping out, and reporting the bug. :heart:

PR #47, opened 2020-07-14
Updated calculateRank.js

garvit-joshi:
I have updated the values and now algorithm will be focusing on a person who has contributed more towardsother people projects. I have also added ```REPPO_OFFSET = 1.10```, and not changed that the value is devided by ```totalRepos```, it helps us to know from when a person is using Git-Hub.


anuraghazra:
Thanks for opening the PR @garvit-joshi ranking algo should be fixed in #49

PR #68, opened 2020-07-15
Change A++ to A+ and A+ to A

Matt-Gleich:
A and S rank should be aligned. So it should either be S++, S+ and A++, A+ or S+, S and A+, A. Is there a reason that this is done in a different way?


anuraghazra:
Hi @Matt-Gleich thanks for the PR,

can you please be more specific why you made this changes? did you made these changes just because the A & S text has to be aligned *visually* ??

Matt-Gleich:
I was more wondering why it is A++ and A+ instead of A+ and A.

Issue #70, opened 2020-07-15
[Feature Request] Vertical Orientation


nathanchu:
There should be a way to change the orientation of the image. For example, [this jsFiddle](https://jsfiddle.net/5ujegnpk/).

filiptronicek:
This looks very good, but I think it has a fairly limited use case. Some things to consider:
* what could be stacked next to it (it would have to be similar height, and if it was another GitHub readme widget, why)
* how it will look, if stacked on mobile.

anuraghazra:
Thanks for requesting the feature @nthnchu i'll see what i can do

nthnchu:
> - what could be stacked next to it (it would have to be similar height, and if it was another GitHub readme widget, why)

I was thinking text could be next to it, so the widget(s) would look like a sidebar. For example:
<hr>
<h1>Heading</h1><img width="228" height="367" align="right" src="https://user-images.githubusercontent.com/63111210/87593750-f29f5880-c6b9-11ea-95c0-fa13c2574a81.png">
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut pharetra sit amet aliquam id diam maecenas. Aliquam sem fringilla ut morbi tincidunt augue interdum velit euismod. Vitae et leo duis ut. Pharetra massa massa ultricies mi quis hendrerit dolor. Tincidunt vitae semper quis lectus nulla at volutpat. Sit amet facilisis magna etiam tempor orci eu lobortis. Aliquam etiam erat velit scelerisque in dictum non consectetur a. Accumsan in nisl nisi scelerisque eu ultrices vitae auctor eu. Tristique risus nec feugiat in fermentum posuere urna nec tincidunt. Nisl nunc mi ipsum faucibus vitae.

[...]

PR #73, opened 2020-07-15
add orientation option to handle vertical and horizontal stats card

omidnikrah:
https://github.com/anuraghazra/github-readme-stats/issues/70


filiptronicek:
@anuraghazra can we please link the issue to this PR?

anuraghazra:
I was thinking if instead of using orientation, we can do "layout" Which would have "vertical" Or "horizontal" Options, that way we can extend out and have more inbuilt layouts in future if we want

omidnikrah:
> I was thinking if instead of using orientation, we can do "layout" Which would have "vertical" Or "horizontal" Options, that way we can extend out and have more inbuilt layouts in future if we want

That's you mean "vertical" and "horizontal" are the acceptable value of "layout" option. Right?

anuraghazra:
Yup `?layout=vertical`

anuraghazra:
Hold on @omidnikrah i'll be merging the PR :D i just wanted to think this layout stuff through, because all these ternary checks and conditionally changing variables seems too much work and look ugly TBH, any ideas?

anuraghazra:
Hmm this does not look right :thinking: 
vertical_bg

Issue #74, opened 2020-07-15
Maximum retries exceeded


klaasnicolaas:
I've tried to set up a deploy in Vercel myself, but immediately from the first moment that I want to request the data via my own url, I receive an error message: **_"Maximum retries exceeded"_**.

Impossible that I could have exceeded my limits right away.

Link that I use: https://github-readme-stats-beryl.vercel.app/api?username=klaasnicolaas&show_icons=true&title_color=fff&icon_color=79ff97&text_color=9f9f9f&bg_color=151515

anuraghazra:
@klaasnicolaas make sure you have a environment variable called PAT_1

rahulrachh:
Could you add to the readme file how to do that?

anuraghazra:
Yes actually sorry the readme's guide is outdated after #58 i'll update it :D

rahulrachh:
I am still confused @anuraghazra.  do we have to put our GitHub generated token in variable PAT_1 instead of GITHUB_TOKEN variable?

anuraghazra:
Yes.... just delete the GITHUB_TOKEN variable, and instead add a token named "PAT_1"

klaasnicolaas:
I changed the variable, it just doesn't change anything. Maybe because I need to update the fork after all today's merges?

anuraghazra:
Update your fork, and redeploy. and make sure that you have the tokens in vercel's env variable section. 

Use PullBot to automatically update your forks, https://probot.github.io/apps/pull/

klaasnicolaas:
Awesome! I got it working 😄 

Thanks, by the way, for the pull probot app tip, which is super handy than that I always have to manually update everything.

rahulrachh:
@anuraghazra it is working 😄 .  Yeah, thanks by the way for the pull bot.

anuraghazra:
Hi @klaasnicolaas i hope your issue is solved, closing this now. feel free to open another one if you have any other queries :D  Thanks!

PR #75, opened 2020-07-16
Added proper apostrophe in stats card's header

sreetamdas:
this PR focuses on two things:

- if the user's name ends with an `s`, no longer add `'s` at the end
- fix an issue where `GITHUB_TOKEN` from the env variables would no longer be respected, instead attempting to use ``process.env[`PAT_${retries + 1}`]`` &mdash; #58

### DIFF
--- a/src/renderStatsCard.js
+++ b/src/renderStatsCard.js
@@ -124,9 +124,11 @@ const renderStatsCard = (stats = {}, options = { hide: [] }) => {
   });
 
   // Conditionally rendered elements
+
+  const apostrophe = ["x", "s"].includes(name.slice(-1)) ? "" : "s";
   const title = hide_title
     ? ""
-    : `<text x="25" y="35" class="header">${name}'s GitHub Stats</text>`;
+    : `<text x="25" y="35" class="header">${name}'${apostrophe} GitHub Stats</text>`;
 
   const border = hide_border
     ? ""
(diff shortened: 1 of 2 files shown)


sreetamdas:
apologies, had to revert the verbose console log message that I'd added; I'll add more test cases for this

anuraghazra:
Hi @sreetamdas thanks for the PR. i was just think that i would probably wont support the GITHUB_TOKEN and only support the PATS since it could be hard to manage later. 

And i was think to just update the readme to solve the issue #74 

I can still merge these changes though, only the "fix: remove 's' after apostrophe if name ends with 's'"

sreetamdas:
alright, sure!
reverted to resolve only the first one

anuraghazra:
@sreetamdas LGTM, can you add a small test to for this?

SirWindfield:
Btw, if the username ends in an `s`, you should still append an `'` after it: `Thomas' Github stats`. The same applies to an `x`: `Felix' GitHub statis`

anuraghazra:
@SirWindfield yup, and i think this would actually do Thomas' 

```js
        <text x="25" y="35" class="header">${name}'${ name.slice(-1) === "s" ? "" : "s"} GitHub Stats</text>
```

But i think the names ends with X has to be considered as @SirWindfield mentioned 

@sreetamdas

PR #76, opened 2020-07-16
Add gh-index field to stats

b1f6c1c4:
gh-index is defined by the h-index of the stargazer count of each of the repo of a GitHub user, i.e. the maximum `h` where there are at least `h` repos with stargazer counts greater than or equal to `h`. This is a good metric of a GitHub user's reputation.

There are tools to calculate this:
- [gh-index](https://github.com/b1f6c1c4/gh-index) on npm
- [github-index](https://github.com/flowmemo/github-index)

TODO:
- [x] Adjust rank based on h-index

### DIFF
--- a/tests/calculateRank.test.js
+++ b/tests/calculateRank.test.js
@@ -5,6 +5,7 @@ describe("Test calculateRank", () => {
   it("should calculate rank correctly", () => {
     expect(
       calculateRank({
+        ghIndex: 4,
         totalCommits: 100,
         totalRepos: 5,
         followers: 100,
@@ -13,6 +14,6 @@ describe("Test calculateRank", () => {
         prs: 300,
         issues: 200,
       })
-    ).toStrictEqual({ level: "A+", score: 49.16605417270399 });
+    ).toStrictEqual({ level: "A+", score: 51.11796226386711 });
   });
 });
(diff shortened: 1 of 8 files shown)


anuraghazra:
Hi @b1f6c1c4 Thanks for opening the PR, 

Can you please explain what this "gh-index" is?

b1f6c1c4:
@anuraghazra It is defined as "the maximum `h` where there are at least `h` repos with stargazer counts greater than or equal to `h`." This concept comes from the concept of `h-index` in the academia, which is well-documented in [Wikipedia](https://en.wikipedia.org/wiki/H-index).

anuraghazra:
@b1f6c1c4  I see, okay this looks good. but i don't want to make the h-index visible to the user's stat in the stats card because most people would get confused about what it is maybe we can just use that to calculate the rank internally.

b1f6c1c4:
Well I think it's OK to hide it by default, but it'd be nice if there could be an option to enable it. IMO people from academia would be more familiar with the concept of h-index.

tests/calculateRank.test.js
```diff
@@ -13,6 +14,6 @@ describe("Test calculateRank", () => {
         prs: 300,
         issues: 200,
       })
-    ).toStrictEqual({ level: "A+", score: 49.16605417270399 });
+    ).toStrictEqual({ level: "A+", score: 51.11796226386711 });
```
Dentrax, line 17: You can consider adding more test cases for each level. (See #98 )

Issue #95, opened 2020-07-18
[Feature Request] Allow `show_icons` to be used with repos' cards


Paveloom:
**Is your feature request related to a problem? Please describe.**
Since most of my repositories don't have any stars and aren't forked by anyone (he-he...), I would like to be able to hide these stats from the repos' cards. This is how standard GitHub's pins behave.

**Describe the solution you'd like**
There's already a `show_icons` parameter for stats' cards, to have something similar to this for repositories' cards (excluding this additional information) would be quite pleasant.

**Describe alternatives you've considered**
This feature can also be added as a new parameter, like `show_stats` or something like that.

**Additional context**

![](https://github-readme-stats.vercel.app/api/pin/?username=paveloom-j&repo=Scats.jl)

anuraghazra:
Oh yeah right, github hides the star/fork if they are zero. i'll probably do that too.

But i don't think i will be introducing any other show_repo_stats parameter for manually showing/hiding the star/fork count because i don't want to bloat the APIs with lots of options, i wanted the Github Extra Pins feature to be a bit more strict in terms on customization.

Paveloom:
Yep, this seems to be right now. Thank you.

PR #97, opened 2020-07-18
Hide star and fork stats in repo card if count of equals to zero

omidnikrah:
https://github.com/anuraghazra/github-readme-stats/issues/95


anuraghazra:
Hi @omidnikrah thanks for the PR, 

As i mentioned in https://github.com/anuraghazra/github-readme-stats/issues/95#issuecomment-660488300 i don't want to introduce the show_icons option in repo card, Probably you can make the automatically hide if the star count and fork count is zero.

omidnikrah:
I agree. I wanted to do the same thing but I think you want to implement an option :)
I will update my PR.

src/renderRepoCard.js
```diff
@@ -52,19 +52,23 @@ const renderRepoCard = (repo, options = {}) => {
...
+        </g>
+      `}
 
-      <g transform="translate(${220 - shiftText}, 100)">
-        <svg class="icon" y="-12" viewBox="0 0 16 16" version="1.1" width="16" height="16">
-          ${icons.fork}
-        </svg>
-        <text data-testid="forkcount" class="gray" x="25">${totalForks}</text>
-      </g>
+      ${totalForks > 0 && `
```
anuraghazra, line 64: I'm not sure how the github pin works exactly, is it like when both are 0 it'll hide the counts or each selectively :thinking:

omidnikrah: As I check repositories in GitHub, it's selectively.

anuraghazra: Okay it's good then!

src/renderRepoCard.js
```diff
@@ -52,19 +52,23 @@ const renderRepoCard = (repo, options = {}) => {
         <text data-testid="lang" class="gray" x="15">${langName}</text>
       </g>
 
-      <g transform="translate(${155 - shiftText}, 100)">
-        <svg class="icon" y="-12" viewBox="0 0 16 16" version="1.1" width="16" height="16">
-          ${icons.star}
-        </svg>
-        <text data-testid="stargazers" class="gray" x="25">${totalStars}</text>
-      </g>
+      ${stargazers.totalCount > 0 && `
```
anuraghazra, line 55: Using && would cause the rendered markup to have a `false` string in it.

svg_false_pr

src/renderRepoCard.js
```diff
@@ -52,19 +52,23 @@ const renderRepoCard = (repo, options = {}) => {
...
+        </g>
+      `}
 
-      <g transform="translate(${220 - shiftText}, 100)">
-        <svg class="icon" y="-12" viewBox="0 0 16 16" version="1.1" width="16" height="16">
-          ${icons.fork}
-        </svg>
-        <text data-testid="forkcount" class="gray" x="25">${totalForks}</text>
-      </g>
+      ${totalForks > 0 && `
```
anuraghazra, line 64: This looks good, but what if a repo has 0 stars and 1 fork? 

Then the design would have a gap in-between the `fork` and the `language` i think

omidnikrah: Right :)))).
I will fix it

anuraghazra: I think i need to implement some kind of `Flex System` to solve this common layout issues

anuraghazra:
LGTM, i've implemented a FlexLayout function which i would commit after merging thing :D Thanks!

PR #100, opened 2020-07-18
refactor: added FlexLayout for flex layouts

anuraghazra:
Added a FlexLayout function to make our life easier with auto layout flows as mentioned in here https://github.com/anuraghazra/github-readme-stats/pull/97#discussion_r456799606 to solve this issue https://github.com/anuraghazra/github-readme-stats/pull/97#discussion_r456799098

But this function as some limitations since the elements are not rendered so i can't get their calculated widths before hand and make a proper flex layout engine. but this works for now.

### DIFF
--- a/src/renderRepoCard.js
+++ b/src/renderRepoCard.js
@@ -1,4 +1,9 @@
-const { kFormatter, encodeHTML, fallbackColor } = require("../src/utils");
+const {
+  kFormatter,
+  encodeHTML,
+  fallbackColor,
+  FlexLayout,
+} = require("../src/utils");
 const icons = require("./icons");
 
 const renderRepoCard = (repo, options = {}) => {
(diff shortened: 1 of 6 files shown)


src/renderRepoCard.js
```diff
@@ -42,6 +47,31 @@ const renderRepoCard = (repo, options = {}) => {
...
 
+  const svgLanguage = `
+    <g transform="translate(30, 100)">
+      <circle data-testid="lang-color" cx="0" cy="-5" r="6" fill="${langColor}" />
+      <text data-testid="lang" class="gray" x="15">${langName}</text>
+    </g>
+  `;
+
+  const svgStars =
+    stargazers.totalCount > 0 &&
```
anuraghazra, line 58: I'm using short-circuit operator instead of ternary because FlexLayout function filters out empty strings

src/renderRepoCard.js
```diff
@@ -62,28 +92,12 @@ const renderRepoCard = (repo, options = {}) => {
...
       <text class="description" x="25" y="70">${encodeHTML(desc)}</text>
       
-      <g transform="translate(30, 100)">
-        <circle data-testid="lang-color" cx="0" cy="-5" r="6" fill="${langColor}" />
-        <text data-testid="lang" class="gray" x="15">${langName}</text>
+      ${svgLanguage}
+      
+      <g transform="translate(${155 - shiftText}, 100)">
+        ${FlexLayout({ items: [svgStars, svgForks], gap: 65 }).join("")}
       </g>
```
anuraghazra, line 99: I'm not putting `svgLanguage` into the FlexLayout's item array because it breaks the layout since i cannot know the svgLanguage's width before hand. I know i can calculate the text length but it does not solves the issue entirely since the FlexLayout then would relay on texts instead of actual element's dimensions

src/renderStatsCard.js
```diff
@@ -1,22 +1,11 @@
...
-  lineHeight,
-  showIcons,
-}) => {
+const createTextNode = ({ icon, label, value, id, index, showIcons }) => {
   const kValue = kFormatter(value);
   const staggerDelay = (index + 3) * 150;
-  // manually calculating lineHeight based on index instead of using <tspan dy="" />
-  // to fix firefox layout bug
-  const lheight = lineHeight * (index + 1);
-  const translateY = lheight - lineHeight / 2;
```
anuraghazra, line 19: Don't need to calculate these anymore because we can just use FlexLayout function to handle this

anuraghazra:
Previews: 

Anurag's github stats

Anurag's github stats


Anurag's github stats

vjspranav (changes requested):
Cannot read properties of undefined (reading 'chunks')

rickstaa:
@vjspranav Thanks for your comment. Could you maybe create a new issue with the problem you are experiencing? It would be great if you could provide some steps to reproduce the error.

PR #102, opened 2020-07-19
:pencil: add Chinese translation

JalenChuh:
add Chinese translation

This project is so great, thanks for your contribution! :bow:

### DIFF
--- a/readme_cn.md
+++ b/readme_cn.md
@@ -1,7 +1,7 @@
 <p align="center">
- <img width="100px" src="https://res.cloudinary.com/anuraghazra/image/upload/v1594908242/logo_ccswme.svg" align="center" alt="Github Readme Stats" /> 
+ <img width="100px" src="https://res.cloudinary.com/anuraghazra/image/upload/v1594908242/logo_ccswme.svg" align="center" alt="Github Readme Stats" />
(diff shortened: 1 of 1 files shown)


readme_cn.md
```diff
@@ -0,0 +1,194 @@
...
+GitHub extra pins allow you to pin more than 6 repositories in your profile using a GitHub readme profile.
+
+Yey! You are no longer limited to 6 pinned repositories.
+
+### Usage
+
+Copy-paste this code into your readme and change the links.
+
+Endpoint: `api/pin?username=anuraghazra&repo=github-readme-stats`
+
```
anuraghazra, line 123: LGTM, You missed this part i think.

anuraghazra:
This is awesome @JalenChuh Thanks!

readme_cn.md
JalenChuh: So sorry, I forgot to explain for it.

I have some puzzle in this part, maybe I know the meaning, but I can't translate it exactly.
So I choose not to translate it.

If you think it's necessary to do that, I will try it. But I think that keep as it is is the better.

anuraghazra: I see, no problem then, LGTM! Thanks!

PR #109, opened 2020-07-20
Update & Improvements on Existing Simplified Chinese Translation of README

TonyTang2001:
For this pulling request, I have updated and improved existing Simplified Chinese translation of README file. Details as follows:
- Added new contents from English README file that hasn't been updated in Simplified Chinese translation of README file.
- Translated contents that were left untouched by last translation.
- Replaced a number of half-width punctuation characters with full-width punctuation characters as following language usage habits. However, characters in commands and codes have been left unchanged per convenience of copying and pasting codes.
- Replaced phrases for more precise translation based on contents.


anuraghazra:
Thanks @TonyTang2001 for the PR :D

PR #128, opened 2020-07-21
feat: added Top languages card

anuraghazra:
NEW FEATURE :tada:

Added a `/api/top-langs` route which would show user's top 5 languages.

DONE:
- Fetching logic
- Rendering logic
- Themes support

TODO: 
- Write tests
- More customization work

### DIFF
--- a/api/top-langs.js
+++ b/api/top-langs.js
@@ -11,6 +11,7 @@ const renderTopLanguages = require("../src/renderTopLanguages");
 module.exports = async (req, res) => {
   const {
     username,
+    hide_langs_below,
     hide_title,
     card_width,
     title_color,
@@ -42,6 +43,7 @@ module.exports = async (req, res) => {
       theme,
       hide_title: parseBoolean(hide_title),
       card_width: parseInt(card_width, 10),
+      hide_langs_below: parseFloat(hide_langs_below, 10),
       title_color,
       text_color,
       bg_color,
--- a/api/top-langs.js
+++ b/api/top-langs.js
@@ -1,11 +1,17 @@
 require("dotenv").config();
-const { renderError, clampValue, CONSTANTS } = require("../src/utils");
+const {
+  renderError,
+  clampValue,
+  parseBoolean,
+  CONSTANTS,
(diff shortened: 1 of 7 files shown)


anuraghazra:
Preview

nombrekeff:
Nice, I like it!

There is one thing I would change though. What about making the label the same color as the progress? Have you tried that?

I don't know how it will look, but maybe it makes it easier on the eye.

nombrekeff:
And what about letting the user decide the number of languages? In my case, I would like to add just the most popular one or two.

anuraghazra:
> Nice, I like it!
> 
> There is one thing I would change though. What about making the label the same color as the progress? Have you tried that?
> 
> I don't know how it will look, but maybe it makes it easier on the eye.

Yup tried that, looks bad with different themes, because every lang has different color and not all of them are vibrant

anuraghazra:
> And what about letting the user decide the number of languages? In my case, I would like to add just the most popular one or two.

I'll add a parameter `hide_langs_below=10` to hide languages which are below 10%

nombrekeff:
> > Nice, I like it!
> > There is one thing I would change though. What about making the label the same color as the progress? Have you tried that?
> > I don't know how it will look, but maybe it makes it easier on the eye.
> 
> Yup tried that, looks bad with different themes, because every lang has different color and not all of them are vibrant

Yup, that's what I thought might happen. Okay no problem, looks good like this

nombrekeff:
> > And what about letting the user decide the number of languages? In my case, I would like to add just the most popular one or two.
> 
> I'll add a parameter `hide_langs_below=10` to hide languages which are below 10%

🙌  I like that approach

anuraghazra:
Test hide_langs_below

nombrekeff:
Looks good to me! Can you share what it looks like with default theme?

anuraghazra:
@nombrekeff comment updated

nombrekeff:
Thanks, it looks great as well!

anuraghazra:
I think this is ready to be merged! :shipit:

PR #123, opened 2020-07-21
Added new theme

divyakelaskar:
### Theme Name : Prussian
ss

PR #129, opened 2020-07-21
Added "prussian" theme to themes/README preview

nombrekeff:
Added prussian theme to themes preview. https://github.com/anuraghazra/github-readme-stats/pull/123

PD: As the cache is 86400 it will take some time until it is updated, I can't think of any good solution at the moment. Apart from making it 0 and then update it to 86400 again.

### DIFF
--- a/themes/README.md
+++ b/themes/README.md
@@ -17,7 +17,8 @@ Use `?theme=THEME_NAME` parameter like so :-
 | `default` ![default][default] |     `radical` ![radical][radical]      |           `merko` ![merko][merko]            |
 | `gruvbox` ![gruvbox][gruvbox] | `tokyonight` ![tokyonight][tokyonight] |        `onedark` ![onedark][onedark]         |
 |  `cobalt` ![cobalt][cobalt]   |  `synthwave` ![synthwave][synthwave]   | `highcontrast` ![highcontrast][highcontrast] |
-| `dracula` ![dracula][dracula] |          `dark` ![dark][dark]          |         [Add your theme][add-theme]          |
+| `dracula` ![dracula][dracula] |          `dark` ![dark][dark]          |      `prussian` ![prussian][prussian]        |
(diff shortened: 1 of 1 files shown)


anuraghazra:
Cache should not matter because the by changing `?theme=` the url is changed so cache should be busted

anuraghazra:
@nombrekeff maybe i'll create a Github action to automatically update the readme with new themes (less manual work)

nombrekeff:
> Cache should not matter because the by changing `?theme=` the url is changed so cache should be busted

I would imagine it would be like that, but when previewing the new theme, it showed the default one though... don't know if it's because of the cache or what exactly.

> @nombrekeff maybe i'll create a Github action to automatically update the readme with new themes (less manual work)

That's cool I'm all up for automation, but if we make people add it when adding a new theme, there should not be that much work from our part. But it could be cool nonetheless! Less work for all 😄

nombrekeff:
@anuraghazra It seems it's being cached actually, It's weird because the query params are different... If I reduce the cache to 0 it renders the correct theme, although with 86400 it shows the default theme.

anuraghazra:
> @anuraghazra It seems it's being cached actually, It's weird because the query params are different... If I reduce the cache to 0 it renders the correct theme, although with 86400 it shows the default theme.

Probably you are missing something, that does not happen. i've checked many times. i also sometime manually bust cache by adding ?v=1 to the url

nombrekeff:
Probably, I will check it later when I come back.

nombrekeff:
Just to know, does it render the new theme for you? https://github.com/anuraghazra/github-readme-stats/tree/master/themes

anuraghazra:
Yes it does

nombrekeff:
Ok, thanks. I think what happened is I tested this theme before deploying it, and it cached that instance

PR #141, opened 2020-07-22
clarify import project PAT name for other lang and remove useless white spaces

watain666:
Found same issue as #122 in other lang readme file.
Modified the image so that `PAT_1` shows as the token name instead of `GITHUB_TOKEN`.
Revmoe some useless white spaces in all readme file.


anuraghazra:
Hi, i guess this pr is stale now, feel free to open a new one.

Issue #140, opened 2020-07-22
[Feature Request] Hide specific languages in Top Languages Card


Jasonnor:
**Is your feature request related to a problem? Please describe.**

Because some languages inherently have more code lines than other languages,
I want to hide these exaggerated language statistics. 😥
Thanks!

[![Top Langs](https://github-readme-stats.vercel.app/api/top-langs/?username=Jasonnor)](https://github.com/anuraghazra/github-readme-stats)

**Describe the solution you'd like**

Maybe add a option like `Hiding individual stats` does, e.g.,
```
&hide=["Jupyter Notebook", "HTML"]
```

**Describe alternatives you've considered**

Not yet 😅

terror:
We currently have the option to hide languages below a certain percentage. 

Either we implement the option to hide languages above a percentage or we add the option to hide specific languages and forget the idea of hiding languages that fall into a percentage category.

Jasonnor:
Yes, as title says, what I want is to hide a specific language 😃

nombrekeff:
I suggested this when adding this feature, but Anurag preferred to add the below percentage option.

I think this would be a nice option to have.

anuraghazra:
> I suggested this when adding this feature, but Anurag preferred to add the below percentage option.
> 
> I think this would be a nice option to have.

I think we can remove the below percentage option maybe? if we are adding specific language hiding features they i don't think hide_langs_below option would be useful anymore.

nombrekeff:
I agree. I think it would be a bit redundant to keep hide_langs_below if we add the hide languages option.

martinRenou:
> Either we implement the option to hide languages above a percentage

This might not make sense for this specific issue actually, because in this particular issue (Jupyter Notebook being a predominant language that the person does not want to see) the percentage for other languages will need to be recomputed.

martinRenou:
Ideally we could remove multiple languages?

terror:
@martinRenou Yeah I didn't think about that. I think hiding specific languages is the way to go.

@anuraghazra I've done some work on this and got this to work over at https://github-readme-stats-69lhfhxhs.vercel.app/api/top-langs/?username=Jasonnor&hide_langs=[%22Jupyter%20Notebook%22]

Let me know what you guys think

anuraghazra:
resolved in #150

PR #150, opened 2020-07-22
Hide specific languages in "Top languages" card

arjunmahishi:
Addressing #140

### DIFF
--- a/api/top-langs.js
+++ b/api/top-langs.js
@@ -3,23 +3,15 @@ const {
   renderError,
   clampValue,
   parseBoolean,
+  parseJSON,
   CONSTANTS,
 } = require("../src/utils");
 const fetchTopLanguages = require("../src/fetchTopLanguages");
 const renderTopLanguages = require("../src/renderTopLanguages");
 
-const parseJSON = (str) => {
-  try {
-    return JSON.parse(str);
-  } catch(err) {
-    return [];
-  }
-}
-
 module.exports = async (req, res) => {
   const {
     username,
-    hide_langs_below,
     hide,
     hide_title,
     card_width,
@@ -52,7 +44,6 @@ module.exports = async (req, res) => {
       theme,
       hide_title: parseBoolean(hide_title),
       card_width: parseInt(card_width, 10),
-      hide_langs_below: parseFloat(hide_langs_below, 10),
       hide: parseJSON(hide),
       title_color,
       text_color,
--- a/src/renderTopLanguages.js
+++ b/src/renderTopLanguages.js
@@ -23,7 +23,7 @@ const createProgressNode = ({ width, color, name, progress }) => {
   `;
 };
 
-const cleanLangName = name => name.toLowerCase().trim();
+const lowercaseTrim = (name) => name.toLowerCase().trim();
 
 const renderTopLanguages = (topLangs, options = {}) => {
   const {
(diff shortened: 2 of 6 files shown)


martinRenou:
Shouldn't it be a change in the request instead?

Changing the request might recompute automatically the other percentages.

For example if I contributed to `60% C++` and `40% JavaScript`, and wanted to hide the C++, I would like it to show that I contributed `100% JavaScript`, not `40%`.

terror:
@martinRenou I'm pretty sure in order to recompute the percentage values, all we have to do is hide the languages before computing the total size

anuraghazra:
> Shouldn't it be a change in the request instead?
> 
> Changing the request might recompute automatically the other percentages.
> 
> For example if I contributed to `60% C++` and `40% JavaScript`, and wanted to hide the C++, I would like it to show that I contributed `100% JavaScript`, not `40%`.

Yes the calculations should be relative to how many langs are shown

api/top-langs.js
```diff
@@ -12,6 +12,7 @@ module.exports = async (req, res) => {
   const {
     username,
     hide_langs_below,
+    hide_langs,
```
anuraghazra, line 15: Let us be consistent with the API and use `hide` instead of hide_langs

api/top-langs.js
```diff
@@ -44,6 +45,7 @@ module.exports = async (req, res) => {
       hide_title: parseBoolean(hide_title),
       card_width: parseInt(card_width, 10),
       hide_langs_below: parseFloat(hide_langs_below, 10),
+      hide_langs,
```
anuraghazra, line 48: Check out how i'm handling arrays 
https://github.com/anuraghazra/github-readme-stats/blob/03f55e809e417abcc0e4b462d52f3c3c52dd1a27/api/index.js#L47

src/renderTopLanguages.js
```diff
@@ -7,7 +7,7 @@ const createProgressNode = ({ width, color, name, progress }) => {
   const progressPercentage = clampValue(progress, 2, 100);
 
   return `
-    <text data-testid="lang-name" x="2" y="15" class="lang-name">${name}</text>
+    <text data-testid="lang-name" x="2" y="15" class="lang-name lang-name-${cleanLangName(name)}">${name}</text>
```
anuraghazra, line 10: Any particular reasons why you are giving each lang a unique CSS class? or just because you wanted to test them? maybe we can just use queryAllByTestId to do that instead.

arjunmahishi: This was just for testing. I'll update the tests to use `queryAllByTestId` instead.

api/top-langs.js
arjunmahishi: I just checked this online. It looks like the standard way of passing an array through url is `?hide=java&hide=html...` This seems to be supported by the current http framework being used also. So, now, I am passing the values like this:
```js
hide: (typeof hide == 'string') ? [hide] : hide,
// ?hide=java -> returns ['java']
// ?hide=java&hide=html -> returns ['java', 'html']
```

arjunmahishi: Do you agree with this? 
Or should we just go with `?hide=['html','java']`

arjunmahishi: Also, if we do `JSON.parse()` on a string that is not in correct json format, the card will not render

anuraghazra: Yes i know the standard way is like that, but i did not used it because i know in future we might wanted hide lots of options and the query string will get very messy and big, that's why i choose the direct JSON string approach also because it ensures the data is in JSON.

arjunmahishi: Done. Updated the PR

anuraghazra:
LGTM Thanks @arjunmahishi 

Note: I've deprecated hide_langs_below option.

martinRenou:
Thanks for the amazing work, this repo deserves every single star it gets ;)

PR #176, opened 2020-07-24
feat: support keep top language

leavesster:
1. add `top` parameter to keep top number languages.
2. add EN and CH README.
3. remove duplicate parameters.


anuraghazra:
Hi @leavesster we already have hiding specific languages option implemented in #150 so hiding top languages is not needed anymore. we previously had "hide_langs_below" option to hide langs below certain threshold but we already deprecated it for the same reason.

leavesster:
I just want to keep top languages no matter what language they are. I know hide specific languages can do it too, I just want to make it more simple, but it's true that top languages changes slow.

PR #191, opened 2020-07-25
draft: codebase refactor

rjoydip:
This PR is contains below list.

- **Refactored** 
    - Entire code base, 
    - Test cases
    - Folder structure
    - File name
- **Improved**
   - code style
   - code format
   - coverage
   - removed redundant code
- **Implement**
    - preact with SSR

etc...

**Note:** This is the draft and initial version of changes. It'll take forward if this gets approved. Fell free to open for discussion.


anuraghazra:
@rjoydip thanks for the PR.
Well this is a big one, you should've opened an issue first and did this progressively. 

I was thinking to do the same refactor, but instead of using react, preact or any server side rendered libs i was thinking to create a custom JSX pragma just for simple strings instead of components and stuff because we don't need all that.

So like a custom `createElement` function which would spit out strings.

rjoydip:
I agree with custom JSX pragma. It will be lightweight and customizable. I choose preact just for lightweight component style.

anuraghazra:
Closing this in favor of #200 & too much conflicts. will have to work on each thing separately to avoid breaking stuff.

Issue #200, opened 2020-07-26
code refactor


rjoydip:
**Is your feature request related to a problem? Please describe.**
> No.

**Refactor the code base**

- [x] Folder structure
- [ ] Simplify code
- [ ] Code style & format [eslint, prettier]
- [ ] Improve code coverage
- [ ] Improve styling
- [x] Removed redundant code
- [ ] Server side rendering

Ref: #200

anuraghazra:
@rjoydip i would say go slow, lets first get the Code Style & Eslint infra ready & merged.

rjoydip:
Correct. That will be better in first try.

PR #207, opened 2020-07-27
Add Eslint infra (#200)

maddyb99:
I think this should check one box for this issue!!
Will try to cover more as and when I get time :smile:


anuraghazra:
Hi @maddyb99 thanks for the PR, but in https://github.com/anuraghazra/github-readme-stats/issues/200#issuecomment-663989910 we are talking about adding esling and associate infra to format code not manual formating of code.

maddyb99:
So, you mean an Eslint config that autformats on save or something?

anuraghazra:
@maddyb99 yeah and also with husky precommit hook

maddyb99:
OK! @anuraghazra will try to do that, so keeping this PR open for now

maddyb99:
There are some issues with styles right now as the linter wants camel case but most places the codebase is using underscore separated words.
Also, I'll test it with husky hook also once locally!

maddyb99:
@anuraghazra is there any specific style being followed here or that has to be decided with the infrastructure being set?

PR #231, opened 2020-07-28
Added Gatsby Theme

paolotormon:
Hi, this theme is inspired by the colors of the popular book/movie "The Great Gatsby"

Hope you like it!

anurag-sample

### DIFF
--- a/themes/index.js
+++ b/themes/index.js
@@ -119,7 +119,7 @@ const themes = {
     text_color: "FFFFFF",
     bg_color: "050F2C",
   },
-  "gatsby":{
+  "great-gatsby":{
     title_color: "ffa726",
     icon_color: "ffb74d",
     text_color: "ffd95b",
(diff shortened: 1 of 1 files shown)


anuraghazra:
Hi @paolotormon thanks for the theme contribution, this looks good but can you provide any references of where did you get the color codes? because this does not really seems like "gatsby" theme

anuraghazra:
Oh its "The Great Gatsby" okay got it.

themes/index.js
```diff
@@ -119,6 +119,12 @@ const themes = {
     text_color: "FFFFFF",
     bg_color: "050F2C",
   },
+  "gatsby":{
```
anuraghazra, line 122: maybe the name "gatsby" can also refer to gatsbyjs so i think it'll be better if we change it to something else. great-gatsby?

paolotormon: Yes, that change would be great!

anuraghazra:
Thanks @paolotormon for the theme contribution. :)

paolotormon:
> Thanks @paolotormon for the theme contribution. :)
You're welcome @anuraghazra ! 😃

PR #249, opened 2020-07-29
refactor: Simplifying rank selection

marcospb19:
Fixing level selection over complicated `if` condition blocks that split into multiple lines using a simpler code with `else if`.

PR #250, opened 2020-07-29
Trailing whitespace cleanup

marcospb19:
When editing code for #249, and saved the file, my code editor removed all the trailing whitespace.

Then I noticed that the project has many laying around.

This is not ideal because it adds to the wrong diff when some file is edited, and it can get confusing when using git's blame functionality.


anuraghazra:
Hi @marcospb19 i think is PR changes gone stale, feel free to open a new one :) closing this for now. 

Sorry for late response.

PR #258, opened 2020-07-29
Changed most used languages height

antoniodecanini:
I only use StatsCard and TopLanguages, so I put them side by side.
The height of the TopLanguages card is less than of StatsCard.
In this PR I changed the height and they were the same.
And it was like this.

e.g

### DIFF
--- a/src/renderTopLanguages.js
+++ b/src/renderTopLanguages.js
@@ -110,7 +110,7 @@ const renderTopLanguages = (topLangs, options = {}) => {
   // RENDER COMPACT LAYOUT
   if (layout === "compact") {
     width = width + 50;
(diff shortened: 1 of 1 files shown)


anuraghazra:
@antoniodecanini thanks for the PR, but we want to have auto adjusted height instead of a fixed height because it would look bad with all that whitespace in the bottom. maybe you can add an option to control the card's height with query param

antoniodecanini:
> @antoniodecanini thanks for the PR, but we want to have auto adjusted height instead of a fixed height because it would look bad with all that whitespace in the bottom. maybe you can add an option to control the card's height with query param

Great @anuraghazra. Good idea control by query param. I'll implements this.

PR #260, opened 2020-07-29
refactor: added reusable Card class to reduce code & test duplication

anuraghazra:
With our current approach we have to handle hide_title, hide_border and auto height adjustment logics with each card and this is causing lot of duplication in code & tests too, thats why i've created a Card.js Class to abstract away the common rendering logic. 

The Card class would automatically handle the hide title, hide border logic as well has when hiding the title it will auto adjust the content & card's height accordingly.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -137,6 +137,7 @@ Customization Options:
 | cache_seconds | number    | manually set custom cache control           | 1800                 | 1800                | 1800                    |
 | count_private | boolean   | counts private contributions too if enabled | false                | N/A                 | N/A                     |
 | layout        | string    | choose a layout option                      | N/A                  | N/A                 | 'default'               |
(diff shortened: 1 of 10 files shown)


readme.md
```diff
@@ -137,6 +137,7 @@ Customization Options:
 | cache_seconds | number    | manually set custom cache control           | 1800                 | 1800                | 1800                    |
 | count_private | boolean   | counts private contributions too if enabled | false                | N/A                 | N/A                     |
 | layout        | string    | choose a layout option                      | N/A                  | N/A                 | 'default'               |
+| card_width    | number    | set the card width                          | N/A                  | N/A                 | 300                     |
```
anuraghazra, line 140: Oooof i forgot to document card_width & i thought it was some git merge conflict issue so i removed it & now i reimplemented it. lol

PR #276, opened 2020-07-31
Added theme

iamrajiv:


anuraghazra:
Hi @iamrajiv thanks for theme contribution.

newtheme

The color contrast isn't good and the blue color is not matching up with dark black color, maybe you can tweak it. 

> NOTE: If you are contributing your theme just because you are using it personally, then you can customize the looks of your card with URL params instead.

iamrajiv:
@anuraghazra I have update the blue color with GitHub contribution  graph color

Issue #287, opened 2020-08-02
Forks Status


Xunzhuo:
**Is your feature request related to a problem? Please describe.**
Can this "Github-Readme-Status" support to summarize the number of "Forks"？

Xunzhuo:
It`s not merged yet, right?

PR #289, opened 2020-08-02
opt-in show stats options

rjoydip:
PR for #287 #327 

Flag: **show_stats=forks**   **Default: null**

tempsnip


WIttyJudge:
🙄🙄🙄

anuraghazra:
Hmm? what is this count?

rjoydip:
Number of repository forked by the user.

rjoydip:
> Please update the Readme to include the "fork" hide option.

Done. Update `readme`.

anuraghazra:
@rjoydip thanks for the PR, the only problem i have with adding more stats is that i don't want to bloat the card with stats and make it longer, so it might be good if we can make them Opt-In instead of Opt-Out

api/index.js
```diff
@@ -61,6 +62,7 @@ module.exports = async (req, res) => {
       hide_title: parseBoolean(hide_title),
       hide_border: parseBoolean(hide_border),
       hide_rank: parseBoolean(hide_rank),
+      show_forks: parseBoolean(show_forks),
```
anuraghazra, line 65: This is not scalable. we have to do something like: show_stats=forks,extra,extra

rjoydip: Okay

rjoydip: done

rjoydip:
removed followers flags https://github.com/anuraghazra/github-readme-stats/issues/327#issuecomment-670471927

Xunzhuo:
why this PR is not merged yet? I think fork status is an important data as well

PR #307, opened 2020-08-04
Add include or exclude languages parameter to top languages api

vn7n24fzkq:
Thanks for your awesome work first.

I add two parameter to include or exclude languages.
Because some framework generate a lot of file which repository owner not actually write those language by self.
So maybe some developer will need this to exclude some language they not good at.

If someone want exclude c++,c# ,Visual Basic, css, html, and javascript.
It would be.
```
http://localhost:3000/api/top-langs/?username={username}&layout=compact&exclude_lang=cplusplus+csharp+visualbasic+CSS+HTML+javascript
```

Because url not allowed some special character, like C++ contains ```+``` or  C# contains ```#```.
So they need replace ```+``` with ```plus``` , ```#``` with ```sharp``` and trim space character.

In my case, I don't write c and c# actually.
```
http://localhost:3000/api/top-langs/?username=vn7n24fzkq&layout=compact&count_private=true
```
image

So I want ignore it.
And after I add exclude language.
```
http://localhost:3000/api/top-langs/?username=vn7n24fzkq&layout=compact&count_private=true&exclude_lang=C+csharp
```
image

If I only want display some language
```
http://localhost:3000/api/top-langs/?username=vn7n24fzkq&layout=compact&count_private=true&theme=dark&include_lang=javascript+java+kotlin+dart+rust
```
image


vn7n24fzkq:
Found an error.
I will push again after fix it.

vn7n24fzkq:
I am enhancing,  make that can include and exclude language

vn7n24fzkq:
Done

tobiasvl:
This would be a great change. To me, the current behavior of hiding languages without replacing them with lesser used languages doesn't make much sense (and yes, I know that @anuraghazra intended it to work like that, but I don't understand why). The main use case for hiding languages must surely be to hide vendored/generated code.

vn7n24fzkq:
@anuraghazra Hi, wondering this pr can be merged, or I need optimize performance.

X3llus:
@anuraghazra I also think this should be merged, I have languages I don't even program with being shown in my top 5 just because of file size/auto generated files from projects.

alexxxdev:
@vn7n24fzkq resolve conflicts, maybe the PR will budge

vn7n24fzkq:
@alexxxdev Thanks.
Theme preview failed, beacuse this PR doesn't change theme.

alexxxdev:
@anuraghazra Will there be any movement on this PR?

hbollon:
Any update on this PR @anuraghazra?

wilmol:
Merge pls!

itsalb3rt:
While this applies, they can exclude languages with special symbols like for example in my case `c++` with the coded the `plus` symbol equivalent `%2b`: `c%2b%2b`

**Normal**:
itsalb3rt	

**Modify**: Exclude c++


itsalb3rt

wilmol:
#445 was recently merged so you can (kind of) achieve this with a combination of `hide` and `langs_count`. But it's not really a robust solution, the languages you hide could one day fall out of the top languages, so you suddenly have more showing on the card.

I'm not sure why @anuraghazra chooses to merge #445 but completely ignore #307.

PR #309, opened 2020-08-05
Added a theme

fernandopal:
Added the `dark-orange` theme


anuraghazra:
Hi @fernandopal thanks for the theme contribution but i think we already have a very similar theme. `great-gatsby`

PR #314, opened 2020-08-05
Add Nordtheme

Algo-ryth-mix:
adding a www.nordtheme.com inspired theme using nord0, nord4, nord8 & nord9
Preview


anuraghazra:
Looks good @Algo-ryth-mix thanks!

PR #329, opened 2020-08-07
Code coverage badge is missing in readme_es.md

qwerty541:


anuraghazra:
Hi @qwerty541 thanks for the PR. looks like i would have to write a script to auto update the common elements in all the readmes.

PR #361, opened 2020-08-11
Update readme_kr.md

jun108059:
### Section link & Relative link does not work

Hello!👋 Thank you very much for making such a good open source.

I found a small problem. Reference links do not work after moving the language-specific guide to the `docs ` folder.

I am sending you `Pull request` as I found it while looking at Korea document.

Thank you :)


anuraghazra:
Thanks @jun108059

Issue #325, opened 2020-08-07
NOTICE: DEPLOYMENT IS CURRENTLY DISABLED


anuraghazra:
Hi everyone, currently github-readme-stats is down because of this ->

![vercel-support](https://user-images.githubusercontent.com/35374649/89615630-a7392e00-d8a4-11ea-9818-443c8b32703f.PNG)

anuraghazra:
Meanwhile consider deploying on your own fork. follow the guide -> https://github.com/anuraghazra/github-readme-stats#deploy-on-your-own-vercel-instance

TanishShinde:
Is there a timeline in which it will be fixed?

anuraghazra:
> Is there a timeline in which it will be fixed?

It should be fixed right now. but now sure if it will happen again.

TanishShinde:
Ok Thanks! 😇

Spiderpig86:
Heads up, seems to be down again.

meganyin13:
Yup, my README stats seem to be down as well

GideonWolfe:
Can confirm!

JubayerJoy:
Still down!

hector-manny:
still down

AbdElraoufSabri:
still an issue

AnmolTomer:
> > Is there a timeline in which it will be fixed?
> 
> It should be fixed right now. but now sure if it will happen again.

Still down. Let us know if you need help with testing or in any way we can help.

J2TeamNNL:
![image](https://user-images.githubusercontent.com/26674308/89700915-a4017900-d95c-11ea-8834-aa1502bfe626.png)
I have deploy my own Vercel but have this issue

chandan-reddy-k:
@J2TeamNNL did you create your own token and save it in vercel environment as described in point 9?

> Select root and keep everything as is, just add your environment variable named PAT_1 (as shown), which will contain a personal access token (PAT), which you can easily create here (leave everything as is, just name it something, it can be anything you want)

ravenCrown0627:
Same here, and I had tried to create my own token and both vercel environment and GitHub too, and it is still not working 

> @J2TeamNNL did you create your own token and save it in vercel environment as described in Readme.md point 9?
> 
> > Select root and keep everything as is, just add your environment variable named PAT_1 (as shown), which will contain a personal access token (PAT), which you can easily create here (leave everything as is, just name it something, it can be anything you want)

[...]

PR #399, opened 2020-08-17
fix: added temporary blacklist

anuraghazra:
Temporarily blacklisting few usernames because these are causing serverless function timeouts.


anuraghazra:
related to #325

PR #422, opened 2020-08-23
refactoring CalculateRank

romadryud:
Just moving var to let and if to switch.

### DIFF
--- a/src/calculateRank.js
+++ b/src/calculateRank.js
@@ -1,15 +1,15 @@
 // https://stackoverflow.com/a/5263759/10629172
 function normalcdf(mean, sigma, to) {
-  let z = (to - mean) / Math.sqrt(2 * sigma * sigma);
-  let t = 1 / (1 + 0.3275911 * Math.abs(z));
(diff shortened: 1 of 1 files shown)


src/calculateRank.js
```diff
@@ -57,34 +57,33 @@ function calculateRank({
...
-  if (normalizedScore >= RANK_A2_VALUE && normalizedScore < RANK_A3_VALUE) {
-    level = "A+";
-  }
-  if (normalizedScore >= RANK_A3_VALUE && normalizedScore < RANK_B_VALUE) {
-    level = "B+";
+  switch (true) {
+    case normalizedScore < RANK_S_VALUE:
+      level = "S+";
+      break;
+    case normalizedScore >= RANK_S_VALUE && normalizedScore < RANK_DOUBLE_A_VALUE:
```
moorereason, line 72: These can all be simplified to a single condition per case:
```javascript
case normalizedScore < RANK_DOUBLE_A_VALUE:
```

PR #436, opened 2020-08-31
docs: add hide_border option in all language docs

weisswolfi:
Add `hide_border` option in all language docs


anuraghazra:
Thanks @weisswolfi

PR #459, opened 2020-09-13
Add more tests for calculateRank

nrchandan:
Also, change let to const for API response.

### DIFF
--- a/tests/calculateRank.test.js
+++ b/tests/calculateRank.test.js
@@ -16,3 +16,35 @@ describe("Test calculateRank", () => {
     ).toStrictEqual({ level: "A+", score: 49.16605417270399 });
   });
 });
+
+describe("Test A++ calculateRank", () => {
+  it("should calculate rank correctly", () => {
+    expect(
+      calculateRank({
+        totalCommits: 1000,
+        totalRepos: 50,
+        followers: 1000,
+        contributions: 610,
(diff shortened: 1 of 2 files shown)


tests/calculateRank.test.js
```diff
@@ -16,3 +16,35 @@ describe("Test calculateRank", () => {
...
+      calculateRank({
+        totalCommits: 0,
+        totalRepos: 0,
+        followers: 0,
+        contributions: 0,
+        stargazers: 0,
+        prs: 0,
+        issues: 0,
+      })
+    ).toStrictEqual({ level: "A+", score: 51.03597635931825 });
```
MindLabor, line 48: I have seen that there are other ranks like B+ and C. This test means that the lowest rank is A+ which was not intended, I think.

nrchandan: True, looks like `calculateRank` is broken. Will wait for input from @anuraghazra

Plavit:
Still an issue

tests/calculateRank.test.js
Plavit: Would it maybe make sense to propose the score to be "0" in this case, with the lowest possible rank (C)?

Plavit (commented):
The rank calculation appears broken, these tests might provide a good incentive to change and improve it

PR #471, opened 2020-09-18
Last Updated x Repositories

ardacetinkaya:
This PR adds new API and new UI card to show last updated X repositories.
It is just a simple fetcher, _repos-fetcher.js_  as repo-fetcher.js

It just fetches all repositories with given username and within _&top=5_ parameters, it is possible to set last updated repositories count.

## Demo
[Repositories](https://github.com/ardacetinkaya?tab=repositories)


Bas950:
Add the hide border option

src/fetchers/repos-fetcher.js
```diff
@@ -0,0 +1,87 @@
...
+            edges {
+                node {
+                  id
+                  name
+                  url
+                  updatedAt
+                  forkCount
+                  stargazers{
+                    totalCount
+                  }
```
Bas950, line 38: Why are you getting forkCount etc. if you dont use it?

ardacetinkaya: Hide border and title option is added and also not required fields for GraphQL request is removed for now.

PR #479, opened 2020-09-19
fix: use consistent case for card titles

max-niederman:
The WakaTIme stats card used different casing for the title than other cards.


anuraghazra:
Hey @max-niederman thanks for the fix, can you please update the test snapshots?

max-niederman:
@anuraghazra I'll get on that.

PR #497, opened 2020-09-26
fix: use consistent case for card titles

max-niederman:
The WakaTIme stats card used different casing for the title than other cards.


anuraghazra:
Thanks @max-niederman

PR #539, opened 2020-10-04
improve: improve translathion of Chinese

jalenchuh:
1. Improve translathion of Chinese
2. Change some words into English

Some words seems not good when translating into Chinese. There is a website like GitHub in China, it also keep some words in origin. So I restored them.


anuraghazra:
hi @jalenchuh thanks for the improvements you also need to change the texts on the test files btw

jalenchuh:
So sorry for that, it's all right now @anuraghazra

anuraghazra:
LGTM! Thanks for the PR @jalenchuh

PR #540, opened 2020-10-04
Added np translations

cimplesid:


anuraghazra:
Cool1 thanks @cimplesid

PR #542, opened 2020-10-04
URL Encoding Help

ajmeese7:
Some people unfamiliar with URL encoding may get confused by this, when this site can just handle it for them.


anuraghazra:
Thanks @ajmeese7

PR #543, opened 2020-10-04
Added missing stats card options in docs translations

qwerty541:
Unfortunately, I cannot guarantee that all of these translations are grammatically correct. But I think that it would be better if these translations would be corrected later than information about these options would be missing.


anuraghazra:
Right we can improve it later on, Thanks @qwerty541

PR #547, opened 2020-10-05
Improve chinese translation

1574242600:


Hephaest (changes requested):
duplicate of **使用方法**

Issue #556, opened 2020-10-06
Add m sign


donno2048:
I think this thing speaks for itself:
<img width="167" alt="Screenshot 2020-10-06 180105" src="https://user-images.githubusercontent.com/61805754/95219369-f7c5fc80-07fd-11eb-9912-d4d329384c21.png">
it would be great to be able to use the m letter for a million or to be able to show to the whole count (or both)

cxong:
https://www.npmjs.com/package/d3-format seems like a nice library to use for this

PR #559, opened 2020-10-06
fixed #555 , #556 issue

### LINKED ISSUE #555: Make smaller when possible
Even when I use `hide_rank=true&hide_title=true` the whole thing stays wide
I think it would be best if the card will get 'thinner' when using `hide_rank=true`

susanta96:
Fixed the card width depending on the hide_rank and hide_title params.


susanta96:
Added The M and With reduced when some params set to hide. Hope It will be Approve.

PR #565, opened 2020-10-09
chore(card-translation): improve Japanese translations

JohnTitor:
It seems some wordings are machine-translated, tweaked them as a native Japanese.


anuraghazra:
Thanks @JohnTitor

PR #570, opened 2020-10-13
Updated translations.js to include Portuguese (Portugal) translations.

joseborges:
Updated translations.js to include Portuguese (Portugal) translations.


anuraghazra:
Hi @joseborges thanks for the contribution.

Although we already have the pt-br 🤔 do we really need the pt-pt? i have no idea.

joseborges:
Hi, I get this question all the time, and I always explain the same, that Portuguese (from Portugal) is not the same as Portuguese (from Brasil), there are several differences, many more than between "American" english and "British" english.

I understand you don't see it this way, since it's a foreigner language to you, but I'm not Brazilian, I'm Portuguese, so my native language is Portuguese and not Brazilian.

That being said, this is a recurring issue, and I accept if you don't merge this PR on your project for, in your opinion, they are the same, but that won't change the fact they are trully diferent.

Kind regards.

joseborges:
Hi, could you please add this PR so that it counts to the #hacktoberfest 2020? It's a legit PR... Thanks!

anuraghazra:
@joseborges ahh I see.

Yup got it. Yeah in our native language we also have different version of same language. 

I'll accept the PR then. Thanks!

PR #576, opened 2020-10-14
Add explanation of the ranks

koppor:
I did not understand the ranks at first sight and did not find "A++" in the issues nor in README.md. After a more closer investigation, I saw the reference to the code. I tried to understand the code and to formulate some small explanation. The result is open sourced in this PR.


anuraghazra:
Looks good thanks @koppor

PR #583, opened 2020-10-19
Greek translation

panosru:
I added a Greek translation (locale `el`)


panosru:
I will make a new PR since I messed this one.

Issue #580, opened 2020-10-17
Rank is off-center


adamalston:
It appears that the rank is off-center both vertically and horizontally relative to the progress circle.

Screenshot
<img height="137px" src="https://user-images.githubusercontent.com/18297826/96355050-80b91f80-10ab-11eb-9e49-8fdfea19dad4.png" />

Live
<img align="" height="137px" src="https://github-readme-stats.vercel.app/api?username=adamalston&hide_title=true&hide_border=true&show_icons=true&include_all_commits=true&line_height=21&bg_color=0,EC6C6C,FFD479,FFFC79,73FA79&theme=graywhite" />

URL: `<img height="137px" src="https://github-readme-stats.vercel.app/api?username=adamalston&hide_title=true&hide_border=true&show_icons=true&include_all_commits=true&line_height=21&bg_color=0,EC6C6C,FFD479,FFFC79,73FA79&theme=graywhite" />`

anuraghazra:
Hi @adamalston thanks for reporting the issue, we will look into it.

related to #55

vSanjo:
Hey @anuraghazra, (and @adamalston!). 
I've pushed a small PR towards fixing this for you. It looks like some minor logic was pulling the information both ways - and I'd argue it isn't required providing the width of the `text` element is big enough. 

Hopefully that helps!

adamalston:
I think this issue should be reopened

PR #602, opened 2020-10-30
Update ranking alignment within dynamic circles.

vSanjo:
It appeared the text was being offset through two sets of logic. One part was centering in all axis, whilst another was off-setting it (or lack of) in anticipation of it being longer or otherwise different.

I've removed that logic (as it's implicit with `alignment-baseline="central"`, `baseline="central"` and `anchor="middle"`) and should look great now!

Such a minor thing, but hopefully it helps with #580.


vSanjo:
I'd really appreciate getting this across the line for Hacktoberfest 2020, if possible!

anuraghazra:
Hi @vSanjo thanks for the PR.

But does not seem centered to me

image

Issue #610, opened 2020-11-02
Names ending on "S"


Th3Ph4nt0m:
**Is your feature request related to a problem? Please describe.**
My Name ends on "S". The title of my card is "Henrik S's stats".

**Describe the solution you'd like**
The title should be "Henrik S' stats"

**Describe alternatives you've considered**
Tried to change the "S" to my full surname, same problem.

harshhhdev:
@anuraghazra can I work on this?

nthnchu:
@Th3Ph4nt0m This is definitely an issue, but in the meantime you can use the custom title query parameter. Try using `&custom_title=Henrik%20S%27%20stats` as a query parameter as a workaround.

anuraghazra:
Ahhh this happened because your name ends with capital S. 

it's a quick fix. 

https://github.com/anuraghazra/github-readme-stats/blob/master/src/cards/stats-card.js#L80

![image](https://user-images.githubusercontent.com/35374649/98150276-64fba900-1ef4-11eb-8e0f-9ac5e2074e17.png)


@harshhhdev yeah you can work on it

nthnchu:
@anuraghazra @harshhhdev I already made the PR. @Th3Ph4nt0m It should work without the workaround once merged.

anuraghazra:
@nthnchu thanks for the fix!

anuraghazra:
closed by #612

PR #612, opened 2020-11-04
fix: names ending in "X" or "S"

nthnchu:
#610
https://github-readme-stats-git-patch-capital-s.nthnchu.vercel.app/

### DIFF
--- a/src/cards/stats-card.js
+++ b/src/cards/stats-card.js
@@ -77,7 +77,7 @@ const renderStatsCard = (stats = {}, options = { hide: [] }) => {
     theme,
   });
 
-  const apostrophe = ["x", "s"].includes(name.slice(-1)) ? "" : "s";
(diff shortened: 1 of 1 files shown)


src/cards/stats-card.js
```diff
@@ -77,7 +77,7 @@ const renderStatsCard = (stats = {}, options = { hide: [] }) => {
     theme,
   });
 
-  const apostrophe = ["x", "s"].includes(name.slice(-1)) ? "" : "s";
+  const apostrophe = ["x", "s", "X", "S"].includes(name.slice(-1)) ? "" : "s";
```
anuraghazra, line 80: I think it would be better if we use

```js
["x", "s"].includes(a.slice(-1).toLocaleLowerCase()) ? "" : "s";
```

PR #633, opened 2020-11-09
Adds pagination support for top languages card

tmathmeyer:
As part of this there is also optional pagination support for pagination
in src/common/retryer.js, so that if you like what this does, it should
be trivial to add support for it to the other fetchers as well. I can
even add them in this PR if you'd like.


tmathmeyer:
Right now anyone with >100 created repos only sees info about the oldest 100, which is really inconvenient.

tmathmeyer:
Any interest in merging this?

PR #646, opened 2020-11-15
Update translations.js

gaetan1903:
Add  mg  locale refer to Madagascar Country


gaetan1903:
When this locale is dispo ? 

Thanks

PR #665, opened 2020-11-26
Response type

### LINKED ISSUE #624: Feature Request: JSON API
**Is your feature request related to a problem? Please describe.**
It would be cool if there was a json api for each card. 

**Describe the solution you'd like**
For example `/api/json?username=nthnchu` would return:
```json
{
  "name": "Nathan Chu",
  "stars": 2,
  "commits_this_year": 561,
  "pull_requests": 18,
  "issues": 52,
  "contributed_to": 15,
  "grade": "A+"
}
```
Yes, I know I could just fetch from the API, but this would simplify things if I wanted to include it on my site, but in a different layout.

**Describe alternatives you've considered**
Using the GH api

**Additional context**
N/A

nathanchu:
Fixes #624

- [x] Response Types
  - [x] JSON
  - [x] XML
  - [x] JSONP
  - [x] YAML
- [x] Cards
  - [x] Stats
  - [x] Top langs
  - [x] Repo
  - [x] Wakatime
- [x] Tests
  - [x] Response Types
    - [x] JSON
    - [x] XML
    - [x] JSONP
    - [x] YAML
  - [x] Cards
    - [x] Stats
    - [x] Top langs
    - [x] Repo
    - [x] Wakatime
- [x] Docs


anuraghazra:
Theme preview (bot)
        Cannot create theme preview

anuraghazra:
Theme preview (bot)
        Cannot create theme preview

anuraghazra:
Theme preview (bot)
        Cannot create theme preview

anuraghazra:
Hi @nathanchu great work on this. 

btw i was wondering that we should actually prevent this feature from direct use from our serverless function because i don't want everyone to overuse these JSON APIs since these json responses will make our serverless functions more generic and lot of people would use it. 

Possible solution is to just add a return statement if users are trying to get any of these responses, and to use this feature they can simply fork the repo, remove the return statement, and deploy on their own vercel instance.

nathanchu:
@anuraghazra I agree. I don't like modifying the source code though so maybe we can let people use an environment variable? My other idea was "access tokens" that could be specified as env vars and used as ?`access_token=user_specified_access_token`.

anuraghazra:
> @anuraghazra I agree. I don't like modifying the source code though because that makes it harder to pull so maybe we can let people use an environment variable? My other idea was "access tokens" that could be specified as env vars and used as `?access_token=user_specified_access_token`.

Okay so you mean in our fetcher function we will use a custom env token to do the fetching, which users would provide (but we won't by default)?

nathanchu:
> > @anuraghazra I agree. I don't like modifying the source code though because that makes it harder to pull so maybe we can let people use an environment variable? My other idea was "access tokens" that could be specified as env vars and used as `?access_token=user_specified_access_token`.
> 
> Okay so you mean in our fetcher function we will use a custom env token to do the fetching, which users would provide (but we won't by default)?

Not a github token, just an unrelated token that is used authenticate the user, if `req.query.access_token === process.env.ACCESS_TOKEN` we can allow the user api access.

anuraghazra:
> Not a github token, just an unrelated token that is used authenticate the user, if `req.query.access_token === process.env.ACCESS_TOKEN` we can allow the user api access.

access tokens does change often so we should consider that... 

also

> Not a github token, just an unrelated token
Which unrelated token? not sure what you are referring to. can you elaborate or maybe add a POC?

anuraghazra:
Theme preview (bot)
        Cannot create theme preview

nathanchu:
> > Not a github token, just an unrelated token that is used authenticate the user, if `req.query.access_token === process.env.ACCESS_TOKEN` we can allow the user api access.
> 
> access tokens does change often so we should consider that...
> 
> also
> 
> > Not a github token, just an unrelated token
> 
> Which unrelated token? not sure what you are referring to. can you elaborate or maybe add a POC?

So a password set in an env var, `API_PASSWORD`, and as a query param, `?api_password`. If `process.env.API_PASSWORD === req.query.api_password`, then api access is allowed.

Ex. I set `API_PASSWORD` env var to `example123`, `/api?username=nathanchu&response_type=json&api_password=example123` returns the card.

PR #709, opened 2020-12-11
wakatime-embed : Customize Date

Instrye:
Wakatime By Embed
https://wakatime.com/share/embed

Can custom date
image

example:

https://vercel.ddnpc.com/api/wakatime-embed?username=instrye&embedCode=2916255c-a02d-4b4b-b7e8-d3533f0fd393
I did not make a readme, hope someone can add.

### DIFF
--- a/src/fetchers/wakatime-embed-fetcher.js
+++ b/src/fetchers/wakatime-embed-fetcher.js
@@ -1,0 +1,22 @@
+const axios = require("axios");
+
+const wakaTimeEmbedData = async ({ username , embedCode}) => {
+    try {
+        const { data } = await axios.get(
+            `https://wakatime.com/share/@${username}/${embedCode}.json`,
+        );
+
+        return data.data;
+    } catch (err) {
+        if (err.response.status === 404) {
+            throw new Error(
+                "Wakatime Data not found",
+            );
+        }
+        throw err;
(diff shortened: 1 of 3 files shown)


src/fetchers/wakatime-embed-fetcher.js
```diff
@@ -0,0 +1,22 @@
...
+        return data.data;
+    } catch (err) {
+        if (err.response.status === 404) {
+            throw new Error(
+                "Wakatime Data not found",
+            );
+        }
+        throw err;
+    }
+};
```
ensia96, line 18:
```suggestion
const wakaTimeEmbedData = ({ username, embedCode }) =>
  axios
    .get(`https://wakatime.com/share/@${username}/${embedCode}.json`)
    .then(({ data: { data } }) => data)
    .catch(err => {
      throw err.response?.status === 404
        ? new Error('Wakatime Data not found')
        : err
    })
```

I think we can save some memory with this code. : )

anuraghazra:
We don't need a whole separate card for embedCode, We can just integrate the parameter in the existing wakatime card.

I'll think about it if we need this and open a separate PR, or you can do it. 

Closing this PR for now since it's got stale a lot. Feel free to open another one.

PR #767, opened 2020-12-31
Fix pt br translations on /src/translations.js

carlos3g:
Some _pt-br_ translations in `/src/translations.js` are wrong, in this PR I'm fixing them.


anuraghazra:
Thanks @carlos3g

PR #770, opened 2021-01-01
Fix readme_pt-BR.md... (All Translations)

rafaelfaustini:
The banner image from Powered By Versel was broken due to invalid relative path on every translated wiki, this pull request fixes all of them.

PR #771, opened 2021-01-01
docs: fix broken links at translated docs

KunHwanAhn:
The translated documents are using broken relative links.

So I fixed it to use right relative links.


KunHwanAhn:
https://github.com/anuraghazra/github-readme-stats/pull/770 PR has similar diffs, but my PR includes it. 🤔

tests/renderStatsCard.test.js
```diff
@@ -245,7 +245,7 @@ describe("Test renderStatsCard", () => {
       document.querySelector(
         'g[transform="translate(0, 25)"]>.stagger>.stat.bold',
       ).textContent,
-    ).toBe("总提交 (2020):");
+    ).toBe("总提交 (2021):");
```
KunHwanAhn, line 248: I'm not sure whether we should check this `year`'s number? 🤔 ...

Anyway, it is the reason why test failed, I fixed.

KunHwanAhn: or we need to write code to check it dynamically? 🤔

anuraghazra:
Thanks for the fix @KunHwanAhn

PR #788, opened 2021-01-06
Fix total commits not counting by year

### LINKED ISSUE #778: Fetching Bug
**Describe the bug**
It's January 2nd, and I don't think I've committed more than 1 time in the past 2 days. I have 666 commits according to the image.

**Expected behavior**
Less than 1 commit for 2021

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**
Live:

Screenshot:
image

**Additional context**
I suspect that's from the last 1 year (January 2, 2020 to January 2, 2021), not just the current year (January 1st, 2021 to January 2nd, 2021)

### LINKED ISSUE #787: Year or number of total commits looks invalid
**Describe the bug**
The `Total commits (2021)` reports value of 241 commits for me. As we have 6th January and I didn't spend last time on massive commiting, it looks like wrong value of commits or eventually the `2021` should be just `last year`.
 
**Expected behavior**
Correct value of commits (just from 2021) or correct year annotation (e.g. `last year`).

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**

obraz

Live demo on [my profile](https://github.com/kam193)

aboqasem:
This fixes #778, fixes #787. But I don't know if the modified code will break something.
I think it should not because `if(!include_all_commits)` means `NOT all commits, but yearly commits`.
Not sure, so feel free to correct me if I'm wrong 😅.

### DIFF
--- a/src/fetchers/stats-fetcher.js
+++ b/src/fetchers/stats-fetcher.js
@@ -61,7 +61,7 @@ const totalCommitsFetcher = async (username) => {
   const fetchTotalCommits = (variables, token) => {
     return axios({
       method: "get",
-      url: `https://api.github.com/search/commits?q=author:${variables.login}`,
+      url: `https://api.github.com/search/commits?q=author:${variables.login} committer-date:>${new Date().getFullYear()}-01-01`,
       headers: {
         "Content-Type": "application/json",
         Accept: "application/vnd.github.cloak-preview",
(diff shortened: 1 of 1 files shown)


src/fetchers/stats-fetcher.js
```diff
@@ -118,9 +118,9 @@ async function fetchStats(
   // normal commits
   stats.totalCommits = user.contributionsCollection.totalCommitContributions;
 
-  // if include_all_commits then just get that,
+  // if NOT include_all_commits then just get that,
   // since totalCommitsFetcher already sends totalCommits no need to +=
-  if (include_all_commits) {
+  if (!include_all_commits) {
```
anuraghazra, line 123: If include_all_commits is true.. we should fetch all of the commits since user joined github.

aboqasem: But that is already done above, isn't it? So the code is fetching the same info again.

After I modified the code, `/api/?username=aboqasem&include_all_commits=true` renders the expected output:

Screenshot 2021-01-08 at 8 42 08 AM

As well as `http://localhost:3000/api/?username=aboqasem`:

Screenshot 2021-01-08 at 8 44 29 AM

Could you explain further what am I missing here?
Thank you!

PR #794, opened 2021-01-09
refactor calculateRank.js

kawarimidoll:
Hello, let me propose some refactorings.

- use `else if` instead of duplicated conditions
- use `const` instead of `var`
- simplify expressions

Thank you for maintaining such a wonderful project.

### DIFF
--- a/src/calculateRank.js
+++ b/src/calculateRank.js
@@ -1,19 +1,16 @@
 // https://stackoverflow.com/a/5263759/10629172
 function normalcdf(mean, sigma, to) {
-  var z = (to - mean) / Math.sqrt(2 * sigma * sigma);
-  var t = 1 / (1 + 0.3275911 * Math.abs(z));
-  var a1 = 0.254829592;
-  var a2 = -0.284496736;
-  var a3 = 1.421413741;
-  var a4 = -1.453152027;
-  var a5 = 1.061405429;
-  var erf =
+  const z = (to - mean) / Math.sqrt(2 * sigma * sigma);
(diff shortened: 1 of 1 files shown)


filiptronicek (approved):
LGTM ✅

src/calculateRank.js
```diff
@@ -67,23 +64,13 @@ function calculateRank({
...
+  } else if (normalizedScore < RANK_A2_VALUE) {
     level = "A++";
-  }
-  if (normalizedScore >= RANK_A2_VALUE && normalizedScore < RANK_A3_VALUE) {
+  } else if (normalizedScore < RANK_A3_VALUE) {
     level = "A+";
-  }
-  if (normalizedScore >= RANK_A3_VALUE && normalizedScore < RANK_B_VALUE) {
+  } else if (normalizedScore < RANK_B_VALUE) {
     level = "B+";
```
Plavit, line 74: Was the B+ intended to be the lowest rank available? I thought there used to be A

Plavit (approved):
Looks good as refactoring, though the raking still makes little sense

src/calculateRank.js
kawarimidoll: Yes, the B+ is the lowest and the A is not available now. There was the rank A in the first version, which has been changed to B+.
(please check the [history for src/calculateRank.js](https://github.com/anuraghazra/github-readme-stats/commits/master/src/calculateRank.js))

This PR is simply a refactoring, so the ranking algorithm has not been changed.

PR #806, opened 2021-01-11
Corrected two typos and changed :- to :

volovikariel:
Corrected two typos, as well as changed  `:-` to `:`. 
I'm unsure what `:-` means, I thought it could signify **here comes an example**, but it was also used in the `readme.md` without providing an particular example. I can add them back in if needed.


filiptronicek (approved):
You caught some good ones @volovikariel!

Issue #815, opened 2021-01-15
For some languages the available options have not been updated


uskey512:
**Is your feature request related to a problem? Please describe.**
For some languages, the available options have not been updated.

**Describe the solution you'd like**
want to keep the parameters available on the documentation up to date.

**Additional context**
The following table summarizes the parameters described in the current document.

<table>
	<tr>
		<th>English</th>
		<th>Français</th>
		<th>简体中文</th>
		<th>Español</th>
		<th>Deutsch</th>
		<th>日本語</th>
		<th>Português Brasileiro</th>
		<th>Italiano</th>
		<th>한국어</th>
		<th>Nederlands</th>
		<th>Türkçe</th>
	</tr>
	<tr>
		<th>Common</th>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
		<td></td>
	</tr>
	<tr>
		<th>title_color</th>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
	</tr>
	<tr>
		<th>text_color</th>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
	</tr>
	<tr>
		<th>icon_color</th>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
	</tr>
	<tr>
		<th>bg_color</th>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
	</tr>
	<tr>
		<th>hide_border</th>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
		<td>o</td>
	</tr>
	<tr>
		<th>theme</th>
		<td>o</td>
		<td>o</td>
[...]

PR #816, opened 2021-01-15
docs : Update readme parameter

uskey512:
- Remove common option for all cards. `hide_border `
Currently, `hide_border` is a valid option for all cards, so there is no need to list it as a separate option.

- Updated Stats Card and Language Card options for Japanese.

( related #815 )


anuraghazra:
Thanks for fixing up the mess @uskey512 :)

PR #817, opened 2021-01-16
Corrected rank normal distribution

### LINKED ISSUE #455: Improve ranking system
**Describe the bug**
I was interested in how the rank was calculated and wanted to know how to rank up. I found out that you use a function called normalcdf to normalize the calculated score. Its basically impossible to get a rank below A+ since normalcdf(0, 231, 7) * 100 returns 51.55 which is in the A+ area. 

The code can be found [here](https://github.com/anuraghazra/github-readme-stats/blob/master/src/calculateRank.js).

#### TODOS (Added by @rickstaa)

- [x] Decide on a new grading system, #2265.
    - [x] Create PR that adds this new grading system.
- [x] Merge https://github.com/anuraghazra/github-readme-stats/pull/2116.
- [x] Merge https://github.com/anuraghazra/github-readme-stats/pull/1186.
- [x] Create PR that shows the top user percentage on hover (see https://github.com/anuraghazra/github-readme-stats/issues/2157).

magneticflux-:
This does NOT `fix` #455 completely, but it corrects the normal distribution function to something that is known to be correct.

In the future, I plan to replace the normal CDF with a power law CDF, which is made easier by migrating the existing normal CDF to a math library right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "emoji-name-map": "^1.2.8",
     "github-username-regex": "^1.0.0",
     "prettier": "^2.1.2",
-    "word-wrap": "^1.2.3"
+    "word-wrap": "^1.2.3",
+    "mathjs": "^9.0.0"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@actions/github": "^4.0.0",
     "@testing-library/dom": "^7.20.0",
     "@testing-library/jest-dom": "^5.11.0",
-    "axios": "^0.19.2",
+    "axios": "^0.21.1",
     "axios-mock-adapter": "^1.18.1",
     "css-to-object": "^1.1.0",
     "husky": "^4.2.5",
--- a/src/calculateRank.js
+++ b/src/calculateRank.js
@@ -1,19 +1,9 @@
-// https://stackoverflow.com/a/5263759/10629172
+const { erf } = require("mathjs");
+const sqrt2 = Math.sqrt(2);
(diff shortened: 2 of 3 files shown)


anuraghazra:
Thanks for working on this @magneticflux- 

Can you show/add more examples on how this change is effected the ranks? that would be nice too see and for further tweaking the values.

magneticflux-:
This PR replaces the approximation of the error function (used to compute the normal CDF) with a more precise approximation that is accurate to at least 18 significant decimal digits over the domain of the function. The details of the new approximation can be found here: https://mathjs.org/docs/reference/functions/erf.html

In practical terms, this PR reduces error in calculating the normal CDF. It reduces error near the mean slightly (as seen in the changed test value), but it reduces error at the extremes more.

Some examples:

```
Input | 30, 25, 5
New   | 0.15865525393145707
Old   | 0.15865526383236372
```
```
Input | 30, 25, 1.4241
New   | 0.12651182223941376
Old   | 0.12651187738346226
```

Again, this PR only addresses problem 1 that I bring up in #455. Problem 2, using an ill-suited probability distribution, is still an issue and I intend to open another PR that migrates to a different distribution (using the Math.js library I introduce in this PR).

magneticflux-:
To fix #455 completely, I'd like to log anonymous statistics for requests to the API so that we can create a histogram plot of their score. That would let us calculate a grade based on the average of all `github-readme-stats` users. Additionally, we could take a random sample of GitHub profiles to calculate and use that if we want to grade based on the average of all GitHub users. That would let us validate the assumption that the score follows a power-law distribution.

PR #657, opened 2020-11-19
added some small improvements.

mmvanheusden:
Some minor changes to readme.md.

PR #841, opened 2021-01-28
docs(translations): some minor improvements

qwerty541:
The same changes as in merged pull request #657 but in the whole project.


anuraghazra:
Thanks @qwerty541

PR #845, opened 2021-01-29
add corner rounding parameter so we can make square cards

tcheinen:
Cards are rounded by default which looks kinda ugly when you have two adjacent cards with a dark theme and a light background.  I added support for a parameter which allows users to control the rounding used by the card (and had it default to the original 4.5).  I added implementation, tests, and readme docs but feel free to ping me if I missed anything.  

image

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -31,7 +31,7 @@ module.exports = async (req, res) => {
     custom_title,
     locale,
     disable_animations,
-    rx,
+    border_radius,
   } = req.query;
   let stats;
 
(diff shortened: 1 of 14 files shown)


filiptronicek (changes requested):
Many thanks for the contribution, @tcheinen, I just have one thing that I think could be tweaked. As you can see [here](https://github.com/anuraghazra/github-readme-stats#stats-card-exclusive-options), all the parameters are pretty memorable and easily self-explainable. I think it would be better to name it something like `border_radius` or `rounding`.

tcheinen:
makes sense!  I renamed it to border_radius

filiptronicek (approved):
LGTM ✅

anuraghazra:
This looks great @tcheinen sorry for the delay.

PR #848, opened 2021-01-30
Update CONTRIBUTING.md

Bondcoder:
The variable name needs to be PAT_1 (or PAT_X up to 8). It throws an error when you use "PAT".


anuraghazra:
Thanks for the fix! @Bondcoder

PR #885, opened 2021-02-17
fix: wakatime compact card bargraph rendering

HaraldoFilho:
This fix the rendering problem on the right end of the bargraph on wakatime compact card, as seen in the pictures below.

Before:
wakatime_card_before

After:
wakatime_card_after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ package-lock.json
 .vscode/
 .idea/
 coverage
+vercel_token
+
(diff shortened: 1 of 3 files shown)


anuraghazra:
Hi @HaraldoFilho would be nice if you could update the snapshots `yarn test -u`

anuraghazra:
Great work! thanks @HaraldoFilho

Issue #932, opened 2021-03-09
Number of repos contributed too is not updating


TanishT:
**Describe the bug**
The number of repos in my profile has been stagnant for a while, despite me contributing to many more repos. 

**Expected behavior**
The repo count should be around 7 - 10. Right now it is just at 5. It has been this way for about 1 week, I thought that it was because of the time it takes for the code to query the github API for repo count, but I doubt it queries less than once a week.

hallzy:
Similar issue. My contributed to says 2 right now, but I have definitely contributed to far more than that.

For me it looks like It is actually displaying the number of projects I have contributed to in the last year, instead of total.

TanishT:
> Similar issue. My contributed to says 2 right now, but I have definitely contributed to far more than that.
> 
> For me it looks like It is actually displaying the number of projects I have contributed to in the last year, instead of total.
> 
> edit:
> 
> I just noticed that the documentation for the API being used for this functionality literally says "A list of repositories that the user recently contributed to."... emphasis on 'recently'.
> 
> So, that would do it. Would be nice to get a full count though

https://github.com/anuraghazra/github-readme-stats/issues/932#issuecomment-797061772

PR #937, opened 2021-03-12
Added new option to get total contributions instead of recent contributions

hallzy:
As mentioned in #932 it seems that the contribution value according to the API documentation is 'recent' contributions.

This PR aims to add a new option to get the total number of repositories that you have contributed to.

I can add the documentation to the README files later. It is late here where I am though so I thought I'd get this up now.

Something to consider though is that this new code as to make several API calls (for me, about 27) because I have to go through all the issues and PRs and commits, and you can only get 100 at a time per API call.

This increases the average load time for me to about 8.2 seconds from 2.5 seconds, and significantly increases the number of API calls, which may become a problem for you.

Perhaps a better solution would be to make a request to github to expose the values I am looking for and make this much simpler and faster.

Anyways, I'll leave it up to you to decide if you even want to add this given the performance hit. It is an option, so it will only be a hit on those that use the option, but still.


hallzy:
The diff in code coverage reported is because I haven't added tests yet... Nor have I added documentation. I figure I'd see if this is even something you want to move ahead with first before adding that.

anuraghazra:
> Something to consider though is that this new code as to make several API calls (for me, about 27) because I have to go through all the issues and PRs and commits, and you can only get 100 at a time per API call.
This increases the average load time for me to about 8.2 seconds from 2.5 seconds, and significantly increases the number of API calls, which may become a problem for you.

Thats the real issue. Vercel will throw error if response takes more than 10seconds and 8.2seconds is too long for that reason, and can cause request to fail. 

So i don't think this is possible right now, we need a better way to fetch the data.

hallzy:
Fair enough. I'll try and find a way to make a feature request for the API to make this faster and simpler.

PR #948, opened 2021-03-17
Add Chinese developers favorite color scheme

acehi:
Add Chinese developers favorite color scheme


anuraghazra:
This seems invalid. 

We don't have any themes called `color-cn-dark`

PR #963, opened 2021-03-23
Rename the ranks to a more legible system

alexanderbock:
The current system looks a bit more like a participation trophy with "B+" being the lowest tier.  This change brings the rank naming down to a more sensible S-F naming scheme with E+F still to be determined


alexanderbock:
I don't know if there is something that is asked of me here?

smart2pet (approved):

rickstaa:
This was implemented in #2762.

Issue #909, opened 2021-02-25
Show 'langs_count' languages even when 'hide' is set


CaptainDario:
When I set the top languages card to show 10 languages but hide G-Code it does not show 10 languages anymore. It shows 9 languages.

It would be nice to have a way that int this case, the next most used language is shown.

I set my most used langauges card to show 10 languages and hide "GCode", "Jupyter" and "tex". Now I only have 7 languages shown.

![Capture](https://user-images.githubusercontent.com/51273483/109142206-e191ec00-775e-11eb-8cbb-59390e20bd5c.JPG)

neo-mashiro:
I have the same issue. When `&langs_count=10` alone is specified, the languages card has exactly 10 entries, like so

![image](https://user-images.githubusercontent.com/29720830/109649744-5469e100-7b2a-11eb-8391-d2da9b601dc1.png)

when `&langs_count=10&hide=Jupyter,html,Roff` is specified, it looks like this

![image](https://user-images.githubusercontent.com/29720830/109649996-a0b52100-7b2a-11eb-9244-d42009ae11bf.png)

I'm hiding three languages here, two of them are in the top 10's list so they consumed two `langs_count` even when they are hidden, the other one is not in the top 10's list so it did not consume `langs_count`, as a result now I have eight effective language slots so only eight languages are shown.

This won't be a problem if the hidden languages fall out of the list so I guess I just need to commit more visible code...

I agree that it'd be nice to not include hidden languages in the `langs_count`, but I think it's better to just increase the `langs_count=10` limit since many people need more than 10.

neo-mashiro:
Why can we only have 10 languages while `&card_height` can be as long as 1000? This appears weird to me.....

I see that Wakatime Stats can have more than 12 languages, shouldn't their layouts be the same?

CaptainDario:
I would still like to have this option.

neo-mashiro:
Yes I think this option is quite useful. Perhaps it's just low-priority at the moment.

rickstaa:
@neo-mashiro Is correct. This feature request makes a lot of sense, but we are currently working on higher priority feature requests and bug reports. Feel free to create a PR for this feature!

rickstaa:
Closing this since I can not reproduce the issue anymore.

```md
[![Top Langs](https://github-readme-stats.vercel.app/api/top-langs/?username=anuraghazra&hide=javascript,html)](https://github.com/anuraghazra/github-readme-stats)
```

[...]

PR #989, opened 2021-04-10
fix: show langs_count langs even when hide is set

florianbussmann:
This fixes issue #909 regarding the endpoint `top-langs`

As shown before by @CaptainDario and @neo-mashiro hiding languages conflicts with setting `langs_count`

### Before: `langs_count` is 6, but only 5 are shown when I hide java
image
### After: Hiding languages is applied before limiting based on `langs_count` to get the correct result
image


anuraghazra:
This looks good, thanks @florianbussmann

Issue #1026, opened 2021-04-27
Wakapi URL (domain with subpath) does not work


dashezup:
**Describe the bug**
There is `api_domain` parameter for wakatime card, but it does not work with custom `wakapi` instances with subpath in the API base URL. When the one self-host wakapi the usual way is reverse proxy the app to a subpath of a HTTPS site.
assume the wakapi instant can be accessed from `https://domain.tld/wakapi/`
api access: `https://domain.tld/wakapi/api/v1/users/{username}/stats/{range}`

neither including `/` (slash) nor `%2F` work.

`https://github-readme-stats.vercel.app/api/wakatime?username=username&api_domain=domain.tld%2Fwakapi&range=last_7_days`
it shows:
```
Something went wrong! file an issue at https://git.io/JJmN9 Cannot read property 'status' of undefined
```

**Expected behavior**
Return the wakatime card when `api_domain` is a URL which contains slash

**Additional context**
maybe add `api_base_url`?

dashezup:
You can try this for testing with my wakapi instance

the public data API works properly
`https://ezup.dev/wakapi/api/v1/users/dashezup/stats/last_7_days`

but the wakatime card does not work
`https://github-readme-stats.vercel.app/api/wakatime?username=dashezup&api_domain=ezup.dev%2Fwakatime&range=last_7_days`

PR #1028, opened 2021-04-27
fix: custom wakatime api url with subpath as api_domain

dashezup:
Allow the value of `api_domain` parameter to be HTML entity encoded string, so that API URL with subpath could be accepted not just domain.

as I mentioned in the corresponding issue, custom wakapi instance usually under a subpath, this change is necessary for it to work.

Fixes anuraghazra/github-readme-stats#1026


dashezup:
With this commit, custom wakapi instance which its api url includes a subpath works

https://github-readme-stats-two-ashy.vercel.app/api/wakatime?username=dashezup&api_domain=ezup.dev%2Fwakapi&range=last_7_days

dashezup:
Close since #1235 is merged.

Issue #1029, opened 2021-04-28
calculate rank total_value mistake


cmougan:
In the file src/calculateRank.js:
 
```
const RANK_S_VALUE = 1;
  const RANK_DOUBLE_A_VALUE = 25;
  const RANK_A2_VALUE = 45;
  const RANK_A3_VALUE = 60;
  const RANK_B_VALUE = 100;

  const TOTAL_VALUES =
    RANK_S_VALUE + RANK_A2_VALUE + RANK_A3_VALUE + RANK_B_VALUE;

```

Should not it be
```
  const TOTAL_VALUES =
    RANK_S_VALUE + RANK_DOUBLE_A_VALUE + RANK_A2_VALUE + RANK_A3_VALUE + RANK_B_VALUE;
```

emily876:
can you assign me this issue?

anuraghazra:
Thanks for opening the issue, closing via https://github.com/anuraghazra/github-readme-stats/pull/1031#issuecomment-882052373

PR #1031, opened 2021-04-29
#1029

cmougan:
Solving issue src/calculateRank.js #1029


anuraghazra:
I thanks for the PR but we are looking into totally changing the stats calculation algorithm. So closing this now.

PR #1091, opened 2021-05-29
feat: threshold for wakatime duration

florianbussmann:
This introduces a parameter called `min_seconds` to wakatime cards which drops languages below the provided threshold

image

### Before
image
image

### After
Providing `min_seconds` = 600 will drop all languages with duration below 10 minutes (600 seconds)
image
image

### DIFF
--- a/api/wakatime.js
+++ b/api/wakatime.js
@@ -30,6 +30,7 @@ module.exports = async (req, res) => {
     range,
     border_radius,
     border_color,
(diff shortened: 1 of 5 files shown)


anuraghazra:
Hi @florianbussmann thanks for the PR. do you think seconds is a good metric? since stats aren't gonna be realtime i think min_minute is fine.

PR #1114, opened 2021-06-05
Apostrophe for English singular possessive is always s

bl-ue:
...even if the name ends with s or x

e.g.

"Anil Das'" is incorrect; "Anil Das's" is correct.
"Felix'" is incorrect; "Felix's" is correct.

For more info, see
https://www.thepunctuationguide.com/apostrophe.html#possessives

### DIFF
--- a/src/cards/stats-card.js
+++ b/src/cards/stats-card.js
@@ -35,10 +35,10 @@ const createTextNode = ({
     <g class="stagger" style="animation-delay: ${staggerDelay}ms" transform="translate(25, 0)">
       ${iconSvg}
       <text class="stat bold" ${labelOffset} y="12.5">${label}:</text>
-      <text 
-        class="stat" 
-        x="${(showIcons ? 140 : 120) + shiftValuePos}" 
-        y="12.5" 
+      <text
+        class="stat"
+        x="${(showIcons ? 140 : 120) + shiftValuePos}"
+        y="12.5"
         data-testid="${id}"
       >${kValue}</text>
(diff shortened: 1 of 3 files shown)


bl-ue:
ping @anuraghazra :)

bl-ue:
@anuraghazra.

rickstaa:
@bl-ue Sorry for the late reply. I think the issue #1114 fixes is quite opinionated, with each style guide having its own rules around the **singular** possessive form  (see https://www.myenglishpages.com/english/grammar-lesson-possessive-form-proper-nouns-ending-s-z-x.php). Therefore, I believe that both `Anil Das's` and `Anil Das'` are correct for the English translation. I, however, agree that `"Felix'` is incorrect and "Felix's" is correct. Some style guides only use an apostrophe for the `x, ch and z` sounds, but these are uncommon. I, therefore, think that if @anuraghazra agrees (he added this code in https://github.com/anuraghazra/github-readme-stats/commit/f7f11d0b2500c0332535e6c3e5ec41819e9d813b), you can remove the `x` from line 97

https://github.com/anuraghazra/github-readme-stats/blob/e9c55d90ad8c9663a480ffaa56484f979ce86fad/src/cards/stats-card.js#L97-L99

For the dutch translation, the current implementation looks correct. There are some additional rules based on the vowel sounds that are hard to implement (see https://www.dutchgrammar.com/en/?n=Pronouns.po06). For the german and Nepali forms, I'm unsure.

PR #875, opened 2021-02-12
fix: renderGradient returns undefined

cuyl:

PR #1133, opened 2021-06-11
fix: return "" instead of return undefined

JacobLinCool:
the function renderGradient returns  undefined if colors.bgColor !== "object"

which may render "undefined" into SVG

and actually, it has already done so


anuraghazra:
Thanks @JacobLinCool

cuyl:
closed #875

PR #1158, opened 2021-06-24
Fix undefined in card svg

tusharsadhwani:
There was an undefined in the svg because of bad return statement, this fixes that.

image


anuraghazra:
Fixed via #1133

PR #1196, opened 2021-07-19
Made Customization navigation better.

OldWizard209:
Customization had loads of parts, so I just improved navigation for all of them.


anuraghazra:
Thanks!

Issue #1206, opened 2021-07-25
Bug(StatsCard): Padding of items seem to be related to the title-length.


Morphclue:
# Describe the bug
I've tried to implement #1155 and noticed, that the padding of the stat-items seems to be related to the title length.

## Actual behavior
"Insgesamt erhaltene Sterne" (see german card for reference) overlaps with the total star count.

## Expected behavior
"Insgesamt erhaltene Sterne" (see german card for reference) should have a padding like "Liczba Gwiazdek dostanych" (see polish card for reference)

#### English
![image](https://user-images.githubusercontent.com/43741877/126891719-385919f2-f622-4c02-9f47-e55a51bd65cf.png)
#### German
![image](https://user-images.githubusercontent.com/43741877/126891742-1dd25f4f-98b6-4b8f-8605-11acc50c4b0e.png)
#### Dutch
![image](https://user-images.githubusercontent.com/43741877/126891789-48eb87a3-a2f7-4321-a8aa-568f8f881d31.png)
#### Polish
![image](https://user-images.githubusercontent.com/43741877/126891926-f8535284-78db-4918-9755-409ee4af3bcd.png)

distrustME:
This bug should be fixed now in #1207. I'll close this issue if the PR gets merged.

Issue #1155, opened 2021-06-23
wrong starts number


532910:
https://github-readme-stats.vercel.app/api?username=532910&show_icons=true&theme=merko

![image](https://user-images.githubusercontent.com/14617699/123138427-aded7700-d45d-11eb-9cd8-a67bce46e6ba.png)

![image](https://user-images.githubusercontent.com/14617699/123138636-ee4cf500-d45d-11eb-91cc-70248246f2f7.png)

John8790909:
@532910 the stars on your profile are the repositories you have stared and the stars in the GitHub stats are the stars people have given your repositories so I think it is displaying the right amount of stars you just go mixed up.

532910:
This way it should be `Total stars received`.

John8790909:
Maybe, but that would be too long don't you think?

532910:
I think the word `Total` should be removed.

John8790909:
Well, then make an issue suggesting that change.

andreakaye:
I wish I had time for this but I work three jobs now because someone went
and fucked up my whole game

GOODBYE

rickstaa:
Closing as this is no longer an issue.

PR #1207, opened 2021-07-25
feat: Added more descriptive name

Morphclue:
# Description
This branch will clarify that the total stars are stars you've received and not given other repositories.
It also fixes the missing padding described in #1206.

### Code
- Changed Translations in `en`, `pl`, `nl` and `de`

## Types of changes
- Enhancement
- Refactor

## Reference
- #1206 
- #1155 

#### Before and After
image
image


src/cards/stats-card.js
```diff
@@ -137,19 +137,6 @@ const renderStatsCard = (stats = {}, options = { hide: [] }) => {
...
-    "es",
-    "fr",
-    "pt-br",
-    "ru",
-    "uk-ua",
-    "id",
-    "my",
-    "pl",
-  ];
-  const isLongLocale = longLocales.includes(locale) === true;
```
anuraghazra, line 151: We actually do want this logic to be present, if we just assume every locale is as long as 100px and shift it by that amount it will look weirdly gaped for short locales. 

So i think we should adjust depending on if the locale is long or not.

distrustME: You are right. I've reverted the commit and added "de" and "nl" to longLocal. This should fix the problem for now.
I think the best solution in the future would be to make the shift dependend on the characters counted for each line and then figure out the padding.
If you agree with me I will create an issue for that so that someone can improve that 👍

anuraghazra:
Thanks @Morphclue

src/translations.js
```diff
@@ -33,16 +33,16 @@ const statCardLocales = ({ name, apostrophe }) => {
       ar: "مجموع النجوم",
       cn: "获标星数（star）",
       cs: "Celkem hvězd",
-      de: "Sterne Insgesamt",
-      en: "Total Stars",
+      de: "Insgesamt erhaltene Sterne",
+      en: "Total stars received",
```
anuraghazra, line 37: cheeky change you put in there, this wasn't suppose to happen lol.

anuraghazra:
Okay thats a mistake, 

Your I got confused by the PR title and the linked issues that did not noticed 

> This branch will clarify that the total stars are stars you've received and not given other repositories.
This change wasn't suppose to happen. 

We are yet to decide on which wording to use. "gained" or "received"

PR #1238, opened 2021-08-10
Add traditional Chinese (zh-tw)

ItzMiracleOwO:
Add traditional Chinese (zh-tw)

### DIFF
--- a/src/translations.js
+++ b/src/translations.js
@@ -6,6 +6,7 @@ const statCardLocales = ({ name, apostrophe }) => {
     "statcard.title": {
       ar: `${encodedName} إحصائيات غيت هاب`,
       cn: `${encodedName} 的 GitHub 统计数据`,
+      "zh-tw": `${encodedName} 的 GitHub 統計數據`,
       cs: `GitHub statistiky uživatele ${encodedName}`,
       de: `${encodedName + apostrophe} GitHub-Statistiken`,
       en: `${encodedName}'${apostrophe} GitHub Stats`,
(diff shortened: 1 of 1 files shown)


src/translations.js
```diff
@@ -84,6 +87,7 @@ const statCardLocales = ({ name, apostrophe }) => {
     "statcard.prs": {
       ar: "مجموع طلبات السحب",
       cn: "拉取请求数（PR）",
+      "zh-tw": "拉取請求數（PR）",
```
rickstaa, line 90: @ItzMiracleOwO Thanks a lot for your contribution. I don't speak traditional Chinese, but the translations look good in google translate. Is there a language syntax reason you added an extra space at the end of the `statcard.prs` and `statcard.commits`. If not, please remove these extra spaces to merge this pr.

rickstaa (changes requested):

src/translations.js
```diff
@@ -58,6 +60,7 @@ const statCardLocales = ({ name, apostrophe }) => {
     "statcard.commits": {
       ar: "مجموع الحفظ",
       cn: "累计提交数（commit）",
+      "zh-tw": "累計提交數（commit）",
```
rickstaa, line 63: See comment below

src/translations.js
ItzMiracleOwO: Hi
Just adding spaces before and after brackets so that the words would seemed seperated on the card, we usually do the same in documents to make it look better

rickstaa: @ItzMiracleOwO Ah, I see. Thanks a lot for the explanation. In that case, the PR looks good to me. Thanks a lot for your contribution.

Issue #330, opened 2020-08-07
What is S rank?


hugovk:
**Describe the bug**

When I add `include_all_commits=true` my badge goes from A++ to S+

**Expected behavior**

I don't know what I expected, but I've no idea what S+ means. S is way down the alphabet, is it much worse than A-F?

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**

```md
[![hugovk's GitHub stats](https://github-readme-stats.vercel.app/api?username=hugovk&count_private=true&show_icons=true&hide_title=true)](https://github.com/anuraghazra/github-readme-stats)
```
[![hugovk's GitHub stats](https://github-readme-stats.vercel.app/api?username=hugovk&count_private=true&show_icons=true&hide_title=true)](https://github.com/anuraghazra/github-readme-stats)

Add `include_all_commits=true`:

```md
[![hugovk's GitHub stats](https://github-readme-stats.vercel.app/api?username=hugovk&count_private=true&show_icons=true&hide_title=true&include_all_commits=true)](https://github.com/anuraghazra/github-readme-stats)
```
[![hugovk's GitHub stats](https://github-readme-stats.vercel.app/api?username=hugovk&count_private=true&show_icons=true&hide_title=true&include_all_commits=true)](https://github.com/anuraghazra/github-readme-stats)

**Screenshot**

![image](https://user-images.githubusercontent.com/1324225/89691122-cf5e7680-d910-11ea-8157-167fefc6a35d.png)

jopemachine:
> I don't know what I expected, but I've no idea what S+ means. S is way down the alphabet, is it much worse than A-F?

I think S means much better than A or A++. (not much worse than A-F).

I think the rating has gone up because the total number of commit is much higher when you include all your commit (private commits).

Metalhearf:
This maybe could interest you, about the _S to F_ grading.
https://en.wikipedia.org/wiki/Academic_grading_in_Japan

hugovk:
OK, thanks, good to know!

I wonder how widely know Japanese grading is. Perhaps it'd be better to shift the grades to begin with A and end with, say, C.

Or add some extra pluse to the existing to A, rather than S.

Thanks!

anuraghazra:
Yeah I guess I'll change this. A lot of people recommend me not to use S because it's confusing. 

I intentionally did this to curve the rankings toward A and A++ even for low active users.

avelino:
[...]

PR #1273, opened 2021-08-30
Changed ranking system

Dosx001:
First off, I would like to discuss about the current ranking system. The ranking system has been discussed numerous times.
#330 #455 #960
And it seems the general consensus is the ranking system is confusing or dissatisfying. 

I'm not trying to be mean or attack with you but I personally I think the rank system is dumb. Specifically with the rank assigned values B+, A, A+, A++, S, and S+. Where's B or S++? My university didn't even give out A+ and here you are giving out A++.

I think there are many ways to improve the current ranking system but you are not willing to implement them. One reason is you don't want people's rank dropping.
https://github.com/anuraghazra/github-readme-stats/pull/960#issuecomment-882050521
And I am assuming you don't want "negative" ranks like F or D.

So I think the best course of action is just to create a completely new system that you will be willing to implement, is not confusing, and is satisfying to the community. One such system would be a numerical rank system. 

Every new GitHub user will start at rank 0 and for every action they make they'll gain points. 
e.g. commits = 30pts, issues = 20pts, follower gained = 5pts, etc.
As users gain points their rank will also go up. Here's how users' rank will be determined. 

Y is the user's total points and and x is their rank. So plug in Y and solve for X. And here's a table showing you some of these X and Y values with the point difference between ranks.

| Rank | Total Points | Point Diff |
|---|---|---|
| 0 | 0 | N/A |
| 1 | 100 | 100 |
| 2 | 215 | 115 |
| 3 | 340 | 125 |
| 4 | 475 | 135 |
| ... | ... | ... |
| 999 | 5,089,900 | 10,085 |
| 1,000 | 5,099,995 | 10,095 |

One reason people dislike the current system is that they believe that they deserve a better rank compared to other users. With this new system, users will be able to distinguish themselves.

This new system will also eliminate confusion, as it is much simpler for our monkey brain to understand. Bigger number, better rank. 🐒🍌

And here it is in action.

I don't expect you to accept this PR, this is the first iteration of the new ranking system. Nothing is set in stone for this ranking system. I do expect you to check it out, provide feedback, and make your own tweaks.

### DIFF
--- a/express.js
+++ b/express.js
@@ -1,15 +1,15 @@
-import statsCard from './api/index.js'
-import repoCard from './api/pin.js'
-import langCard from './api/top-langs.js'
-import wakatimeCard from './api/wakatime.js'
-import express from 'express'
-import dotenv from 'dotenv'
+import statsCard from "./api/index.js";
+import repoCard from "./api/pin.js";
+import langCard from "./api/top-langs.js";
+import wakatimeCard from "./api/wakatime.js";
+import express from "express";
+import dotenv from "dotenv";
 
-dotenv.config()
-const app = express()
-app.listen(process.env.port || 9000)
+dotenv.config();
+const app = express();
+app.listen(process.env.port || 9000);
 
-app.get('/', statsCard)
-app.get('/pin', repoCard)
-app.get('/top-langs', langCard)
-app.get('/wakatime', wakatimeCard)
+app.get("/", statsCard);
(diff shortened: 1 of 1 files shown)


iskandarreza:
Yo what does this even mean? I was at an A+ and I didn't really care but today it says I'm B and now I want answers! lol

PR #1284, opened 2021-09-02
Added prefs theme

Patitotective:
Added prefs theme in reference of ***PREFS*** python library https://patitotective.github.io/PREFS/


Patitotective:
I did read the contribution guidelines, where says:
> NOTE: If you are contributing your theme just because you are using it personally, then you can customize the looks of your card with URL params instead.

Never says when to add the theme, just when don't do. Anyways I'm currently using the [customization options](https://github.com/anuraghazra/github-readme-stats#customization).
Thanks for the answer, i will close the pull request.

PR #1321, opened 2021-09-25
Add Vietnamese Language

nobiit:


src/translations.js
```diff
@@ -158,6 +163,7 @@ const statCardLocales = ({ name, apostrophe }) => {
       sk: "Účasti",
       tr: "Katkı Verildi",
       pl: "Udziały",
+      vi: "Đã Đóng Góp"
```
rickstaa, line 166: @nobiit Thanks for your translation. I don't speak Vietnamese, but I'm reviewing your pull request using google translate. I was just wondering why you're using this over `Góp phần`, which google translate suggests is?

src/translations.js
```diff
@@ -240,6 +248,7 @@ const langCardLocales = {
     sk: "Najviac používané jazyky",
     tr: "En Çok Kullanılan Diller",
     pl: "Najczęściej używane języki",
+    vi: "Ngôn Ngữ Thường Sử Dụng",
```
rickstaa, line 251: Similar here, google translate gives `Các ngôn ngữ được sử dụng phổ biến` which is probably wrong but just double-checking.

rickstaa (changes requested):

rickstaa:
@nobiit Thanks for your pull request. Overall it looks good to me there were two points I wanted to double-check (see my comments).

src/translations.js
nobiit: `Contribute` means `Đóng góp`, and I use `Đã Đóng Góp` to refer to past contribute. `Góp phần`, when translate into English, is `Contribute`, but this word is not commonly used to refer to contribute in my country

src/translations.js
nobiit: - `Phổ biến` is a word that refers to something that is commonly used by people, And I think it's not needed in this case
- `Thường sử dụng` to refer to something commonly used by 1 person, I think this word is more literal

nobiit:
Thanks @rickstaa

Issue #1396, opened 2021-10-19
Cannot read property 'filter' of undefined


Mo-Shakib:
**Describe the bug**
Everything was fine, I did not make any change to my README.me, but yesterday I got this error.
My code:
`<img height="180em" src="https://github-readme-stats.vercel.app/api/wakatime?username=shakib&langs_count=8&layout=compact&theme=dark&hide_border=true&hide_title=true&hide=other">`

**Expected behavior**
Everything was working file yesterday.
**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**
![image](https://user-images.githubusercontent.com/50780268/137848309-135d8a11-144b-48dd-bdae-990f809a1451.png)

**Additional context**

PR #1403, opened 2021-10-21
Fixed bug where wakatime api returns undefined languages

Marcholio:
Fixes: #1396 

For some reason the wakatime API returns "undefined" for languages in this particular case. This change prevents error in that case and normal "No coding activity this week" message is shown. Also added a test case to validate this.

The particular example case also has hide_title=true and layout=compact, so the card will be just empty. Without those it will be like this:
Screenshot 2021-10-21 at 16 53 53


anuraghazra:
Thanks for the fix. :)

Issue #1416, opened 2021-10-24
Self hosted vercel instance gives maxDuration warning


rickstaa:
**Describe the bug**

When I try to deploy the master branch on my self-hosted vercel instance (i.e. https://github.com/anuraghazra/github-readme-stats#deploy-on-your-own-vercel-instance) I get the following error:

```text
The value for maxDuration must be between 1 second and 10 seconds, in order to increase this limit upgrade your plan: https://vercel.com/pricing
```

I did some debugging and this error has been present since 81b6593.

**Expected behaviour**

This error should not be present.

#### TODO

- [x] We should document why this is happening.

PR #1461, opened 2021-11-16
Change maxDuration

w3bdesign:
If maxDuration is higher than 10 it will not allow deployment with the free plan.

Lowering this value will allow people to deploy it.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -2,7 +2,7 @@
   "functions": {
     "api/*.js": {
       "memory": 128,
-      "maxDuration": 30
+      "maxDuration": 10
     }
   },
   "redirects": [


rickstaa:
@w3bdesign Thanks for your pull request. This is a known issue that has been discussed in https://github.com/anuraghazra/github-readme-stats/issues/1416. We can not lower the `max_duration` as it is needed for people on the pro plan (including the vercel instance that is deployed by this repository).

>It's not a bug, it was intentionally reduced to 30s because by default on PRO plan it sets to 1min which is too high and was causing too much memory usage. but for non-pro users the duration can only be <10s thus 30s fails.

@anuraghazra I think we can close this pull request.

Issue #1426, opened 2021-10-28
Toggle bold off or on for stats card


anubhav217:
**Describe the bug**

**Expected behavior**

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**

**Additional context**

anuraghazra:
Duplicate of #1169

PR #1468, opened 2021-11-22
feat(stats): add 'not_bold' option to stats card

rickstaa:
Implements #1426. @anuraghazra I'm not sure if we want to implement this feature in the `main` branch but I wanted to give people a pull request they can use might they want this feature.


api/index.js
```diff
@@ -25,6 +25,7 @@ module.exports = async (req, res) => {
     title_color,
     icon_color,
     text_color,
+    not_bold,
```
anuraghazra, line 28: Really confusing boolean. 

Rename it to "text_bold" and keep it enable by default. if user wants to disable bold they can pass text_bold=false

rickstaa: @anuraghazra I did it like this since you do not specify default values for the options in the README. I, however, i: agree that `not_bold` is even less clear.

rickstaa:
**Text_bold=undefined**

undefined

**text_bold=false**

not bold

**text_bold=true**

not bold

rickstaa:
In the current form, the title is kept bold. We can however also change that behaviour. @anubhav217 did not provide a detailed explanation in #1426.

rickstaa:
Given that the header is implented here:

https://github.com/anuraghazra/github-readme-stats/blob/ecd7d98a3c5ab9f50706458065a314b2c4f7f98e/src/common/Card.js#L153

I don't thing this is a good idea.

rickstaa:
@anuraghazra I think this can be safely merged.

Issue #1134, opened 2021-06-11
Wrong text formatting on Firefox 


DantasB:
**Bug description**
When i see my github stats some part of the text get overrided by other: 
![image](https://user-images.githubusercontent.com/15825133/121753275-70492f80-cae8-11eb-9502-b29215d55ebf.png)

**Expected behavior**
The text should be equal to the google chrome one:
![image](https://user-images.githubusercontent.com/15825133/121753314-835bff80-cae8-11eb-8718-69515308d7da.png)

Issue #664, opened 2020-11-26
Total Commits Rendering Bug


sci4me:
![](https://i.imgur.com/o97eZdh.png)

sci4me:
I am only experiencing this on Linux (5.9.8-arch1-1) in Chromium (86.0.4240.198) and also in Firefox (83.0).

ZerdoX-x:
![image](https://user-images.githubusercontent.com/49815452/101284914-5f032b80-3804-11eb-9711-cb8b10403366.png)
Same.
5.9.11-arch2-1
Mozilla Firefox 84.0b7
Chromium 87.0.4280.88

sci4me:
This issue isn't stale, the bug still occurs.

sci4me:
This issue isn't stale, the bug still occurs.

hasufell:
Same here. I see it everywhere basically. Is there a quickfix, like hiding the year, for example?

sci4me:
Still not stale ffs.

sci4me:
Please forgive me for bumping but, I'm wondering if this is ever going to be fixed? I'm curious to know how many people are experiencing this issue; clearly not enough to make it a priority.

sci4me:
Still not stale.

Electrenator:
Currently also seeing this within the latest Firefox. Next to this the card title appears to be overflowing with the `hide_rank` setting set to true. As can be seen within the picture below :)
![Screenshot 2021-12-09 at 10-26-37 Electrenator - Overview](https://user-images.githubusercontent.com/18311389/145369783-3bd2968f-231b-434d-9d96-615552f6978f.png)

As far as I've seen, every chromium based browser I've tried seems to render the formatting without any problems.

Electrenator:
It looks like that Firefox handles `font-weight` differently than chromium. The following Firefox specific css can be used to fix this. (applied in Electrenator/github-readme-stats@f82f70934e3380b3a1d15fc5a4bb49b1d0e240bf)
```css
@supports(-moz-appearance: auto) {
    .stat{ font-size:12px; }
    .header{ font-size: 15.5px }
}
```
But the current testing method doesn't support the `@supports` selector, which currently makes 15 tests fail. Unsure how to fix that but the cards should work.

sci4me:
This bug is still occurring on my end, on Arch still. Is this in prod yet?

Electrenator:
[...]

PR #1492, opened 2021-12-09
Fixed Firefox text overflow issue

Electrenator:
This pull request fixes the Firefox overflow issues linked within this description. Both the tests and the cards should work like intended but now without text overflow issue within Firefox. ~~Tests currently don't work because it appears that `@supports` is not supported by cssToObject used within the tests. I currently have no idea how to fix that test but the cards themself should work.~~

Fixes anuraghazra/github-readme-stats#664

Fixes anuraghazra/github-readme-stats#1134


anuraghazra:
@Electrenator thanks for the PR, can you try using this cssToObject library instead? https://www.npmjs.com/package/@uppercod/css-to-object

It should work.

Issue #1536, opened 2022-01-16
`Total Commits` has wrong value


anc95:
**Describe the bug**
I found `Total Commit` is wrong for my account

If I visit https://github-readme-stats.vercel.app/api?username=anc95, I got
![image](https://user-images.githubusercontent.com/13167934/149651803-62b6d00a-5a29-4edf-9e3e-9f223c98a25b.png)
`Total Commits` for 2022 is 136

https://github-readme-stats.vercel.app/api?username=anc95&include_all_commits=true
![image](https://user-images.githubusercontent.com/13167934/149651773-203f4558-7c57-4711-8b9f-38d9eaed260c.png)
`Total Commits` for all history is 123

**Expected behavior**
`Total Commits` should display right value

MagicLike:
Duplicate of #1234

Issue #1234, opened 2021-08-09
`include_all_commits` decrease the stats


wiryadev:
**Describe the bug**
`include_all_commits` decrease the stats instead of increasing it.

**Expected behavior**
I expect `include_all_commits` to be increase my stats since it should counts total commits.

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**
Here is the table showing with and without `include_all_commits`
<table style="width:100%">
  <tr>
    <th>With</th>
    <th>Without</th>
  </tr>
  <tr>
    <td><img src="https://user-images.githubusercontent.com/26060382/128685642-a8c04e93-38e7-42db-a445-0f7ed0033a49.png"></td>
    <td><img src="https://user-images.githubusercontent.com/26060382/128685558-757bf24f-9d90-4c78-9af8-7dab3ebb3690.png"></td>
  </tr>
</table>

viticis:
I ran into the same problem.

AyushShahh:
I have the same problem

wy-luke:
same

AlessioQuercia:
In my case this happens only when both ```include_all_commits=true``` and ```count_private=true``` are used.

|  | ```count_private=false``` | ```count_private=true``` |
| :---: | :---: | :---: |
| ```include_all_commits=false``` | ![immagine](https://user-images.githubusercontent.com/18620868/142222761-78cd4f8b-ff06-4317-bf71-10b6f0b8f31b.png) | ![immagine](https://user-images.githubusercontent.com/18620868/142225159-9a801677-85e2-468a-9e07-aa8683378c65.png) |
| ```include_all_commits=true``` | ![immagine](https://user-images.githubusercontent.com/18620868/142223123-1dd7d46f-dbe4-4286-a8a8-308912e7dd9d.png) | ![immagine](https://user-images.githubusercontent.com/18620868/142225447-1eb81535-87b7-478e-9e0a-c57e95315f98.png) |

Perhaps the issue is when both are included?

Furthermore, I just noticed that by adding an additional ```&``` after the github name increases the commits (probably close to the right number, I am not sure, but the ones before are definetely not the right ones):
```
https://github-readme-stats.vercel.app/api?username=USERNAME&&include_all_commits=true&count_private=true
```
Result (using the double ```&```): 
![immagine](https://user-images.githubusercontent.com/18620868/142227125-4386e8e0-7f79-4023-bcb3-e91561ab1591.png)

[...]

PR #1537, opened 2022-01-16
fix: pass start_time when query userinfo

anc95:
fix: #1536 #1234

### DIFF
--- a/src/fetchers/stats-fetcher.js
+++ b/src/fetchers/stats-fetcher.js
@@ -11,11 +11,12 @@ const fetcher = (variables, token) => {
   return request(
     {
       query: `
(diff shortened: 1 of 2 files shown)


anc95:
From [Github API DOC](https://docs.github.com/en/graphql/reference/objects#user), `ContributionsCollection` has `from` and `to` params stand for the contribution range

> `from (DateTime)`
> Only contributions made at this time or later will be counted. If omitted, defaults to a year ago.
>to (DateTime)
> Only contributions made before and up to (including) this time will be counted. If omitted, defaults to the current time or one year from the provided from argument.
```

If you don't set from param, I will be a year ago, which is unexpected at our case

PR #1540, opened 2022-01-17
Update link view demo in pt-BR

victtorfreitas:
This link don't work, because the pt-BR not is demo but is 'demonstração'


rickstaa:
@victtorfreitas Thanks for your contribution!

PR #1564, opened 2022-02-01
update ReadMe change link params from ? to &

NorthernSantan:
Update typo in ReadMe, update parameters for Hide individual languages and Exclude individual repositories from 
?EXTRA_PARAM to &EXTRA_PARAM in accordance with usage and other examples.

When new users use the parameters, merging this request will allow them to avoid an error in their link from copy and pasting existing code.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -266,7 +266,7 @@ Endpoint: `api/top-langs?username=anuraghazra`
 
 ### Exclude individual repositories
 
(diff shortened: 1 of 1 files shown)


rickstaa:
@NorthernSantan Thanks for creating this pull request. I will close it since f7ddcbe757ae9aa6a54f24adb4db7dc821b51be3 fixed this issue.

PR #1577, opened 2022-02-09
Update index.js

mathcodes:


mathcodes (commented):
passed tests, added electron scheme :)

PR #1623, opened 2022-02-28
feat: experimental css inject

anuraghazra:


anuraghazra:
Nah bad idea. too much flexibility & potential to exploit CSS.

PR #1636, opened 2022-03-08
add purple gradient theme

yechukim:
I added purple gradient theme :) 

below is preview 
스크린샷 2022-03-08 오후 3 22 27


rickstaa:
You appear to be changing an existing theme. Can you please explain why you want to change an existing theme?

PR #1644, opened 2022-03-14
feat: gradient units to userSpaceOnUse

unickhow:
Hi all,
I'm building an UI generator for stats cards recently, and I noticed that `bg_color` with gradient result is unexpected in most rotate angles, I think the point is `gradientUnits` attribute, so issue this PR for set it to `userSpaceOnUse` if its ok.

image


anuraghazra:
Thanks @unickhow

PR #1652, opened 2022-03-17
Add extra language colours as defined in WakaTime

TurnrDev:


anuraghazra:
Fixed by #1654

TurnrDev:
An even better solution, amazing!

PR #1654, opened 2022-03-17
Development Script to Automate Language Color Retrieval

Davoleo:
I've written a simple script that populates `src/common/languageColors.json` automatically from [Github's Linguist Repository](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml)

Advantages of using Github Colors:
- Github Colors are generally more populated and almost always updated faster than Wakatime colors
- Updating the colors doesn't require anymore than running the script file (instead of relying on PRs to edit the json file) [unless Github decides to change the format of the yaml file, which is very unlikely]
- Unified color system with other types of cards (no color conflicts)
- This is ultimately a project to display development stats on github so I think it makes sense for it to show Github language colors

_If anything needs to be changed, please tell me and I'll do so_

New Dependencies: 
- [js-yaml](https://github.com/nodeca/js-yaml)
  - Since the original file is in yaml format we need something to parse it into a regular JS object and then write it down in JSON format
  - This is set to be a `development` dependency, as it's only used in a development script

Other Dependencies:
- axios (already a dependency of the project)

Example Comparing Screenshots:
image
image


NightyLighty (approved):

anuraghazra:
This looks great! Thanks @Davoleo

PR #1714, opened 2022-04-23
update README: counting the private contributions

alirezadizaji:
I have updated README a little bit to indicate besides `count_private=true`, publicizing the private contributions is also required.


qwerty541 (commented):
Outdated since #2736 was merged.

PR #1724, opened 2022-04-27
chore: replace `git.io` links

SukkaW:
All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Replace the git.io link with its actual URL.

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -18,7 +18,7 @@ const renderError = (message, secondaryMessage = "") => {
     .gray { fill: #858585 }
     </style>
     <rect x="0.5" y="0.5" width="494" height="99%" rx="4.5" fill="#FFFEFE" stroke="#E4E2E2"/>
-    <text x="25" y="45" class="text">Something went wrong! file an issue at https://github.com/anuraghazra/github-readme-stats</text>
+    <text x="25" y="45" class="text">Something went wrong! file an issue at https://tiny.one/readme-stats</text>
     <text data-testid="message" x="25" y="55" class="text small">
       <tspan x="25" dy="18">${encodeHTML(message)}</tspan>
       <tspan x="25" dy="18" class="gray">${secondaryMessage}</tspan>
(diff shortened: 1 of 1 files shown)


rickstaa:
LGTM. However, the git.io link will still be available as a read-only link. @anuraghazra, what do you think? Do you want to keep the old link since it still works or do you want to migrate to another URL shortener (e.g. https://bitly.com/ and https://short.io/?

SukkaW:
> I read the changelog, and it looks like the git.io link will still be available as a read-only link.

GitHub describes in a way that only git.io links in print documentation and research papers will retain services. Even if they have changed their mind, it is not guaranteed that the `git.io` will stop providing service in the future.

rickstaa:
@SukkaW Your right. I looked over those words. As I think there was a reason for shortening the URL, let's wait for @anuraghazra to decide on a new URL shortener.

anuraghazra:
@rickstaa short.io looks good but seems like I have to login right? Lets use this 

https://tinyurl.com/readme-stats

anuraghazra:
git.io was such a damn good URL tho. It's shame they deprecated it.

src/common/utils.js
```diff
@@ -18,7 +18,7 @@ const renderError = (message, secondaryMessage = "") => {
     .gray { fill: #858585 }
     </style>
     <rect x="0.5" y="0.5" width="494" height="99%" rx="4.5" fill="#FFFEFE" stroke="#E4E2E2"/>
-    <text x="25" y="45" class="text">Something went wrong! file an issue at https://git.io/JJmN9</text>
+    <text x="25" y="45" class="text">Something went wrong! file an issue at https://github.com/anuraghazra/github-readme-stats</text>
```
rickstaa, line 21:
```suggestion
    <text x="25" y="45" class="text">Something went wrong! file an issue at https://tiny.one/readme-stats</text>
```

rickstaa: @SukkaW I think this can be merged if the URL is changed.

rickstaa (changes requested):

rickstaa:
@anuraghazra Yea looks like short.io is a premium service. I can only create URLs if I already own a domain. `tiny.one/readme-stats` looks good. I, however, could not find the monthly click limit. The documentation states the following:

image

I assume that the `-` means that URLs have unlimited link clicks, and the limits given on the premium plans are related to the tracking behaviour.

rickstaa:
@SukkaW Thanks again for your PR! I have merged it into the main branch.

PR #1731, opened 2022-05-02
Added Table of Contents to ReadMe

Tomiiwa:
I added a Table of Contents to make navigation easier.


rickstaa:
@Tomiiwa, thanks for your PR. Although the header is a bit misleading and the TOC and README could use a update, we already have a TOC under the https://github.com/anuraghazra/github-readme-stats#features section.

Issue #1600, opened 2022-02-18
[Proposal] New Top Language Detection Algorithm 


kitswas:
**Is your feature request related to a problem? Please describe.**

This is related to an [earlier issue](https://github.com/anuraghazra/github-readme-stats/issues/481).  
I am not quite happy with the way Top Languages are calculated (just counting the bytes).

**Describe the solution you'd like**

The new method is inspired by the [h-index](https://en.wikipedia.org/wiki/H-index?wprov=sfla1). 

For a language L,  
let `x = byte count`(currently in use) and `y = number of repositories using L`.  
```
x*y = k^2
k = sqrt(xy)
```
Now we can use this k-metric to find the top languages and their weightage.
The higher the value of `k`, the more used is the language.

While this method is an improvement over both byte count and the number of repositories, it is in no way a magic wand.

**Describe alternatives you've considered**

None at the moment.

**Additional context**

I have gone through [this issue](https://github.com/anuraghazra/github-readme-stats/issues/136) mentioned in the FAQ.

_P.S.: This is the first time I am trying to contribute to open source. Please point out if I make a mistake._

anuraghazra:
Hi @kitswas thanks for the proposal, quite interesting. 

What do you think of the downsides of this approach? Can you mention any examples? 
And also how exactly this method is better than just counting bytes? Will be better to have a comparison.

kitswas:
> What do you think are the downsides of this approach? 

The size of a program to perform the same task varies when written in different languages. This metric does not take it into consideration.  
The alternatives (counting bytes and [number of repositories with the language](https://ionicabizau.github.io/github-profile-languages/)) are no better though.

> And also how exactly this method is better than just counting bytes?

When a language appears in multiple repositories, it means that the developer has some degree of reliance on it. This is an attempt to factor it into the calculation.

> Will be better to have a comparison.

Here's a theoretical comparison.
[Comparsion.pdf](https://github.com/anuraghazra/github-readme-stats/files/8102608/Comparsion.pdf)

rickstaa:
Just for reference, in case people want to chip in the discussion relating to the algorithm used to determine the 'top languages '. This issue is also discussed in the related PR #1732.

rickstaa:
This was implemented in https://github.com/anuraghazra/github-readme-stats/pull/1732.

PR #1732, opened 2022-05-03
New top language algorithm implementation

kitswas:
Old Card
========
[Old card](https://github.com/kitswas)

New Card
========
[New card](https://github.com/kitswas)

The tests have been updated.

I am not very familiar with Javascript so I made only minimal changes.  
Some refactoring might be required.

Implements #1600


kitswas:
The deployment URL is  
https://kitswas-github-readme-stats.vercel.app/

Feel free to try and see how the new percentages differ from the old.

kitswas:
There was an error in my earlier implementation.

See the differences below:

- Has error - [Error](https://github.com/kitswas)
- Current - [Current](https://github.com/kitswas)

rickstaa:
@anuraghazra, I agree with @kitswas that the language counting algorithm could be improved. Due to some pressing deadlines, I did not yet have time to go deep into the proposed calculation. I will therefore give you my first thoughts.

From my first impression, the `y` variable (i.e. the number of repositories) influences the percentages. By running an optimisation on the `p` and `q` factors, we might get better results, but I'm afraid that judging the accuracy of the result might be very subjective. Currently, I am inclined to go with the old behaviour since it is similar to what GitHub uses in their repo language overview (i.e. https://github.com/github/linguist/). I, however need to dive in deeper to give a detailed conclusion.

kitswas:
The influence of `y` is felt only when it is greater than `1`, i.e. when a language appears in multiple repositories. 
The numbers `p` and `q` could be optimised. Allowing the users to specify these might be a possibility.
The lower the exponents, the more stable the percentages are.

rickstaa:
Without diving deeper into the change. If we decide to implement this option the following would make sense to me:
- Rewrite the `k = x^p * y^q` in a way that allows us to disable the new behaviour by default.
- By default disabling the influence of `y`.
- Giving users the ability to change the `p` and `q` values.
- Documenting this new behaviour in the README.

src/fetchers/top-languages-fetcher.js
```diff
@@ -84,20 +86,34 @@ async function fetchTopLanguages(username, exclude_repo = []) {
...
+          count: repoCount,
         },
       };
     }, {});
 
+
+  Object.keys(repoNodes)
+    .forEach((name) => {
+      // comparison index calculation
+      repoNodes[name].size = Math.sqrt(repoNodes[name].size * repoNodes[name].count);
```
rickstaa, line 114: Just a general note. The formula used here seems to be different from the formula you gave in https://github.com/anuraghazra/github-readme-stats/issues/1600#issuecomment-1046056305.

````
k = x^p * y^q
```

I think it still uses an old calculation.

```
x*y = k^2
k = sqrt(x*y)
```

In https://github.com/anuraghazra/github-readme-stats/issues/1600#issue-1143665672 you mention that this calculation is based on the h-index but other than the goal that you and this index have in mind I can not see similarities. If you are proposing this form in our case I don't think an `sqrt` is needed since the number of bytes that come out of Graphql will always be possitive.

rickstaa (changes requested):

rickstaa:
Please be aware that I did not do any research on metrics that could give a better 'used languages' representation. I, therefore, can only comment on the metric you proposed in #1732 and #1600. It could be that there are better statistical indexes for solving the problem.

src/fetchers/top-languages-fetcher.js
kitswas: [The new method is inspired by the h-index](https://github.com/anuraghazra/github-readme-stats/issues/1600#issue-1143665672) - I thought of this while reading that [Wikipedia page]((https://en.wikipedia.org/wiki/H-index?wprov=sfla1)). I decided to err on the side of caution and reference that.

We get `k = sqrt(x*y)` when we use `p=0.5` and `q=0.5` in the general formula `k = x^p * y^q`.

kitswas:
The new commits -

- [x] Rewrite the k = x^p * y^q in a way that allows us to disable the new behaviour by default.
- [x] By default disabling the influence of y.
- [x] Giving users the ability to change the p and q values
- [ ] Documenting this new behaviour in the README.

src/fetchers/top-languages-fetcher.js
```diff
@@ -84,20 +86,34 @@ async function fetchTopLanguages(username, exclude_repo = []) {
...
+          count: repoCount,
         },
       };
     }, {});
 
+
+  Object.keys(repoNodes)
+    .forEach((name) => {
+      // comparison index calculation
+      repoNodes[name].size = Math.pow(repoNodes[name].size, p) * Math.pow(repoNodes[name].count, q);
```
rickstaa, line 114: I don't think this is correct as setting q to `0` will now break the algorithm and not fall back to the default behaviour (i.e. `size^1*count^0=0`).

rickstaa (changes requested):

rickstaa:
@kitswas, thanks a lot for creating the new commits. It shows you are devoted to improving this project! I think people like you make a strong community found behind this project! 🚀 

I, however, have to say I'm still not 100% on board with this. When I created https://github.com/anuraghazra/github-readme-stats/pull/1732#issuecomment-1120180189 I overlooked the case where `p` is `0` and `q` is `1`. I don't think this version makes sense with the conception people might have with the language card. Namely, that is similar to the one shown by GitHub. I, however, understand that I'm also biased since I'm happy with the old behaviour and, therefore, likely would not use #1732 myself. I changed #1600 to a proposal so we can get more feedback from the community.

@anuraghazra, what do you think? Another way to implement this would be to create a switching variable `rank_method`, allowing users to switch from the old behaviour to the new behaviour. In that case, we can fix the `p` and `q` to more optimal values. However, this will not make everybody happy and wants more control over the language card and gives us more optimisation work. I, therefore, think there are two paths: keep the old behaviour since it is more consistent with what people expect to see or add the behaviour where people can change the `p` and `q` values.

src/fetchers/top-languages-fetcher.js
kitswas: `a^0 = 1`, not `0`. (For `a != 0`)

kitswas: See line 106 in the [tests](https://github.com/anuraghazra/github-readme-stats/commit/a30cebea9ebb5a20d83dd4afddfaca97d35434bc).

rickstaa: Ah of course! I'm clearly not awake enough to review this PR 💤 . But that is out of the way the PR now looks good to me.

rickstaa:
@kitswas For now, #1732 looks good. Let's wait on what @anuraghazra says. I will review this once more if we decide to merge it.

rickstaa (approved):

kitswas:
Rebased this branch onto anuraghazra/github-readme-stats/master.

kitswas:
Pulled latest changes from [anuraghazra/github-readme-stats/master](https://github.com/anuraghazra/github-readme-stats/tree/master).

rickstaa (approved):

rickstaa:
> count_weight

Sounds like a plan! If you commit these changes, both in the code and the README, I will merge the pull request :).

kitswas:
@rickstaa Updated.

rickstaa (approved):

rickstaa:
Great! Merged into the main. Thanks for your contribution.

PR #1734, opened 2022-05-04
Update docs to mention maxDuration to 10

VivekNeel:
I forked the project and tried running it, but got an error saying 

The value for maxDuration must be between 1 second and 10 seconds, in order to increase this limit upgrade your plan: https://vercel.com/pricing

- We have to set the maxDuration to 10 for free plans in vercel.json


rickstaa:
@VivekNeel Good catch! We forgot to update this when we added this information to the README.md. I have merged this in the master branch.

PR #1735, opened 2022-05-06
Updated Year

Dan0xE:


rickstaa:
@Dan0xE Thanks a lot for your pull request. I however don't think this update is necessary. To my knowledge, the MIT license is valid from the year it first appeared. The year that is shown is the year that the copyright was added (see https://opensource.stackexchange.com/questions/1522/what-should-be-written-in-mit-license-year-full-name). I am however no lawer so feel free to convince me 🤓.

rickstaa:
Based on that reasoning

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Anurag Hazra
+Copyright (c) 2022 Anurag Hazra
```
rickstaa, line 3:
```suggestion
Copyright (c) 2020 Anurag Hazra
```

rickstaa (changes requested):

LICENSE
rickstaa: Based on that reasoning this should be the preferred change (i.e. 003e5e245689905c9141129a4ff8f380a7fdfe37).

rickstaa:
@anuraghazra Are you okay with changing this to 2020 again, or do you want to keep the data updated? Some repositories do this, but they specify a range of 2020-2022. This, however, is redundant.

PR #1778, opened 2022-05-21
Add the option to hide card backgrounds

YuKitsune:
Paired with the right colours, this would allow for cards to blend into the readme a bit better when changing from light to dark mode.


YuKitsune:
It looks like this can be achieved using the theme context tag: https://github.com/anuraghazra/github-readme-stats#use-githubs-theme-context-tag

I'll go ahead and close this off for now.

PR #1783, opened 2022-05-23
japanese translation fix

MrGreensWorkshop:
I checked in Safari, Chrome, Firefox it was all same, "Contribute" 's , Japanese translation was too long to so text was overlap the numbers. its fixed by changing shorter word.

Screen Shot 2022-05-23 at 13 05 58


rickstaa:
Merged into master. Thanks for your contribution!

PR #1833, opened 2022-06-28
Update index.js

import023:


rickstaa:
It looks like your trying to change an existing theme. If this was the intention please explain why this theme should be changed.

PR #1861, opened 2022-07-11
Circleci project setup

Drakunal:


rickstaa:
@Drakunal thanks for your contribution. I'm however not sure why we need a hello world Ci action. 🤔

PR #1903, opened 2022-07-29
ci: change theme-preview action run condition

rickstaa:
This PR makes sure that the theme-preview action only runs when a theme was added or changed.


.github/workflows/preview-theme.yml
```diff
@@ -7,6 +7,8 @@ on:
       - master
       - theme-preview-script
```
rickstaa, line 8: @anuraghazra Why do you include the `theme-preview-script` and `themes/index.js` branches in the `pull_request_target`? Where do we use these?

PR #1954, opened 2022-08-17
fix: test passes with more specific querySelector

### LINKED ISSUE #1851: Test action fails
**Describe the bug**

After 1975116b4b97e8ee11773fae1cd31a46cb1c73ea and 1975116b4b97e8ee11773fae1cd31a46cb1c73ea were merged the ` `tests/card.test.js` now fails. This is strange since these commits do not change any of the code that is tested. See https://github.com/anuraghazra/github-readme-stats/runs/7215331182?check_suite_focus=true.

**Expected behaviour**

The test should pass (See https://github.com/anuraghazra/github-readme-stats/runs/7020934858?check_suite_focus=true). @anuraghazra, did you encounter this problem before? It seems that the latest release of [ubuntu-latest](https://github.com/actions/virtual-environments/blob/releases/ubuntu20/20220626/images/linux/Ubuntu2004-Readme.md) breaks the test. The [stop](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/stop) element inside [lineargradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient) is no longer found by [querySelector ](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) method.

qimolin:
Resolves #1851

### DIFF
--- a/tests/card.test.js
+++ b/tests/card.test.js
@@ -168,18 +168,18 @@ describe("Card", () => {
       "fill",
       "url(#gradient)",
     );
-    expect(document.querySelector("defs linearGradient")).toHaveAttribute(
+    expect(document.querySelector("defs #gradient")).toHaveAttribute(
       "gradientTransform",
       "rotate(90)",
     );
     expect(
-      document.querySelector("defs linearGradient stop:nth-child(1)"),
+      document.querySelector("defs #gradient stop:nth-child(1)"),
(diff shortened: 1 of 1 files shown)


rickstaa:
@qimolin LGTM. Thanks a lot for creating this fix and contributing to the GRS! 🚀🎉

PR #1965, opened 2022-08-24
Fix NaN and undefined for empty profiles

adil192:
On new/empty GitHub profiles, the card would previously appear like this with `include_all_commits` enabled...
```
https://github-readme-stats.vercel.app/api?username=USERNAME&include_all_commits=true
```
before
And now it has the same behaviour as without `include_all_commits`... 
after


rickstaa:
@adil192 Very nice catch! LGTM.

rickstaa:
Merged into the master branch. Thanks for your contribution 🚀.

PR #1983, opened 2022-08-27
Added Finnish Documentation

mazkdevf:
**What is this?**
Hello, This is a Finnish translation on docs.

**A quick idea summary**
My quick idea of the summary is that this project should have many language docs as possible.

**What & why do you want to add the specific feature**
So basically wanted to do this since I really like this project and I saw that it didn't have a Finnish translation so as I know Finnish why wouldn't I try to include it in this project? So here I'm at trying to make this better.

**How the management of this project could check the words?**
I think you guys can use a Translator or this one **https://www.translate.com/english-finnish**

**Picture of Finnish translated on docs**
image

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Anurag Hazra
+Copyright (c) 2022 Anurag Hazra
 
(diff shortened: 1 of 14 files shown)


rickstaa:

PR #2015, opened 2022-09-11
Add cocoa default color

AlfinIndrawan:
cocoa default color is gray which is ugly in any theme especially for those who work Ios dev. (first commit i accidentally delete html but restored again)


rickstaa:
@AlfinIndrawan Thanks for your PR. We however do not generate the ` file ourselves. We fetch it from the https://github.com/github/linguist/ repository using the https://github.com/anuraghazra/github-readme-stats/blob/master/scripts/generate-langs-json.js script. It is therefore best that you add a feature request at https://github.com/github/linguist/discussions?discussions_q=Cocoa for them to add the cocoa language.

PR #2029, opened 2022-09-17
fix: improve boolean parse behaviour

rickstaa:
This commit ensures that incorrectly typed properties like `hide_border=trsdfaf` use their default values.


rickstaa:
@anuraghazra, I made a very small change to the parseBoolean function to prevent confusion.

src/common/utils.js
```diff
@@ -60,17 +60,23 @@ function isValidHexColor(hexColor) {
...
+ * @param {string | boolean} value
+ * @returns {boolean | undefined }
  */
 function parseBoolean(value) {
-  if (value === "true") {
-    return true;
-  } else if (value === "false") {
-    return false;
-  } else {
+  if (typeof value === "boolean") {
```
anuraghazra, line 67: What's this line meant to do?

rickstaa: @anuraghazra, thanks for reviewing this PR! I added that condition since the tests were failing. I, however, understand that it is a bit confusing given the type definition says that the function only takes strings. Alternatively, I can change all the `true` parameters in the tests to true strings.

rickstaa: The function now implements he following behaviour:
- `true` -> `true`
- `false` -> `false`
- true, True, TRUE -> `true`
- false, False, FALSE -> `false`
- Random string -> `undefined`

rickstaa:
@anuraghazra I improved the syntax a bit so the function behaviour is more clear.

rickstaa:
@anuraghazra, I think this PR is good to go.

PR #2054, opened 2022-09-25
ci: add e2e vercel test action

rickstaa:
This PR adds several end-to-end tests that can be used to test whether the Vercel Preview deployment successfully returns the cards. This will prevent broken deployments like 107f7ca52c3431bf989e7c3624ad130e98cb4f5a from ending up on the master branch (see https://github.com/anuraghazra/github-readme-stats/actions/runs/3122948590/jobs/5065233718). @anuraghazra It is similar to what we discussed yesterday. I now use simple calls for testing the cards, but we can add more properties. Let me know what you think 😄.


tests/e2e/e2e.test.js
```diff
@@ -0,0 +1,45 @@
...
+  let VERCEL_PREVIEW_URL;
+
+  beforeAll(() => {
+    VERCEL_PREVIEW_URL = process.env.VERCEL_PREVIEW_URL;
+  });
+
+  test("retrieve stats card", async () => {
+    expect(VERCEL_PREVIEW_URL).toBeDefined();
+    await expect(
+      axios.get(`${VERCEL_PREVIEW_URL}/api?username=willianrod`),
```
anuraghazra, line 19: It could still return but return error. We should test if it's actually returning svg with proper stats.

rickstaa: @anuraghazra Good point. I updated the code a bit so that it now checks the snapshot. While coding, I realised this would break if the stats of the used user accounts (i.e.`anuraghazra` and `willianrod`) are updated and if the code creating the cards is updated. 

I could create a dummy GitHub account in which the stats stay static; however, in that case, the test will still fail when the card generation code is updated. The best solution I could think of now is to generate the cards based on local data and create a static user account for which this data does not change. A bit similar to the regular tests, but then we compare the server SVG with the locally generated SVG.

https://github.com/anuraghazra/github-readme-stats/blob/fe1ca87e33cfbfd75b058fe52bfe7c3e15e2c8c4/tests/renderStatsCard.test.js#L25

Maybe you have a better idea? 🤔

rickstaa: @anuraghazra I implement the solution above, which seems to work (see https://github.com/anuraghazra/github-readme-stats/actions/runs/3142651570/jobs/5106496325).

anuraghazra (approved):
LGTM

PR #2085, opened 2022-10-01
Fix Give India Logo in readme.md

ianbromwichuos:
Replaced .svg that has a dead img url with a JPEG from the official give India website.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -64,7 +64,7 @@
 
 <p>
 <a href="https://indiafightscorona.giveindia.org">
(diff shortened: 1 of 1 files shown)


rickstaa:
@ianbromwichuos Thanks for your PR. I, however, don't think this is needed since the image is shown using the current URL. I think it is simply protected against direct access in the browser.

<a href="https://indiafightscorona.giveindia.org">
Give india logo
</a>

```html
<a href="https://indiafightscorona.giveindia.org">
<img src="https://d2wvdrxmr8p0wf.cloudfront.net/static/giveindia.svg" alt="Give india logo" width="200" />
</a>
```

PR #1186, opened 2021-07-13
Ranking System v2

francois-rozet:
Hello @anuraghazra :wave:

As #883, #455 and #1029, I've noticed that the function `calculateRank` doesn't work as expected.

The first problem is the use of a normal distribution instead of an [exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution), as pointed out by #960 . The second is that each metric has the same weight in the score, while some are clearly better indicators (e.g. stars). I've corrected both problems in this PR.

If you want to try it out before merging, I've hosted this version on Vercel.

Importantly, in #960, being extremely good in a single metric (e.g. 100k commits) while being terrible elsewhere still gets an "S+" rank. In my implementation, in order to get "S+", you need to be outstanding everywhere. However, you can still get "A+" or "S" if a single metric is poor (e.g. contributions or issues) but you have a lot of stars and followers.

PR #2089, opened 2022-10-02
Lower ranking system from `A++` -> `A+` and `A+` -> `A`

CaedenPH:
Going from `A+` to `B+` for a very small amount of contributions seems a bit extreme. I'm open to opinions on the matter, but in my opinion this change makes the ranking system a bit more reasonable.


rickstaa:
@CaedenPH, I understand your frustration. This is a known problem with the current ranking method (see https://github.com/anuraghazra/github-readme-stats/issues/455). There are two PR open to solve this (i.e. https://github.com/anuraghazra/github-readme-stats/pull/960 and https://github.com/anuraghazra/github-readme-stats/pull/1186). We will likely merge one of them in the future.

PR #2104, opened 2022-10-03
Ts migration base

HwangTaehyun:
* [Add typescript and webpack settings](https://github.com/anuraghazra/github-readme-stats/commit/a23a25d9da9f98eddff4d0d731b68505e9959982)
* [Exclude index.js in linguist stats](https://github.com/anuraghazra/github-readme-stats/commit/3809ffed8a6bf5a2e75513de4748f0c76be21542)
* [Add ts-jest settings](https://github.com/anuraghazra/github-readme-stats/commit/2dcc0bf165434fd8e38f96875591fd9b73c06bd3)

I applied typescript & webpack settings to this project and also added ts-jest and enabled test code. In this PR, I focused that typescript migration can be applied partially because this project has many source codes. 

@rickstaa @anuraghazra @GuriX @lostgirljourney Could you review this PR?

PR #2108, opened 2022-10-04
feat: add initial typescript migration

rickstaa:
This commit performs the base changes that are needed to migrate the codebase to typescript.


anuraghazra:
Btw we shouldn't bundle the code, Vercel supports TS out of the box.

rickstaa:
> Btw we shouldn't bundle the code, Vercel supports TS out of the box.

Your right. That's why I closed #2104. Here I use babel for the tests, not for the deployment. Is this also not needed?

https://github.com/anuraghazra/github-readme-stats/blob/bcda6ce01e58cf2be245a57ccf923d3e81518fcc/babel.config.js#L1-L6

anuraghazra:
Oh cool then.

src/common/Card.ts
```diff
@@ -1,7 +1,78 @@
-import { getAnimations } from "../getStyles.js";
-import { encodeHTML, flexLayout } from "./utils.js";
+import { getAnimations } from "../getStyles";
+import { encodeHTML, flexLayout } from "./utils";
 
-class Card {
+/** Card colors. */
+export interface CardColors {
+  /** Title color. */
```
anuraghazra, line 6: All these comments are quite repetitive, the property itself is very self explanetory.

src/common/Card.ts
```diff
@@ -10,7 +81,7 @@ class Card {
    * @param {string?=} args.customTitle
    * @param {string?=} args.defaultTitle
    * @param {string?=} args.titlePrefixIcon
-   * @param {ReturnType<import('../common/utils.js').getCardColors>?=} args.colors
+   * @param {ReturnType<import('./utils').getCardColors>?=} args.colors
```
anuraghazra, line 84: We should remove jsdoc type defs if we are using TS.

src/common/icons.ts
```diff
@@ -1,4 +1,7 @@
-const icons = {
+
+export type Icons = {[key: string]: string};
```
anuraghazra, line 2: Just use a Record<string, string>

src/common/utils.ts
```diff
@@ -1,15 +1,16 @@
...
+import { themes } from "../../themes/index";
+import type { ThemeEnum } from "../../themes/index";
 
 /**
  * @param {string} message
  * @param {string} secondaryMessage
  * @returns {string}
  */
-const renderError = (message, secondaryMessage = "") => {
+const renderError = (message: string, secondaryMessage = "") => {
```
anuraghazra, line 13: We set message to string but we didn't set secondaryMessage to string. In parameters we should be explicit about the types.

src/common/utils.ts
```diff
@@ -275,26 +297,30 @@ const SECONDARY_ERROR_MESSAGES = {
...
-  constructor(message, type) {
+  constructor(message: string, type: keyof typeof SECONDARY_ERROR_MESSAGES) {
     super(message);
     this.type = type;
     this.secondaryMessage = SECONDARY_ERROR_MESSAGES[type] || type;
   }
 
-  static MAX_RETRY = "MAX_RETRY";
-  static USER_NOT_FOUND = "USER_NOT_FOUND";
+  static MAX_RETRY = "MAX_RETRY" as "MAX_RETRY";
```
anuraghazra, line 312: `as const` doesnt work?

Issue #1177, opened 2021-07-08
The most used languages only count the oldest 100 projects


tc-imba:
**Describe the bug**

The most used languages count only count the oldest 100 projects sorted by time. The new projects are not counted

**Expected behavior**

The most used languages count should count all projects

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**

I test it in https://docs.github.com/en/graphql/overview/explorer

The current graphql is like

```graphql
{
  user(login: "tc-imba") {
    repositories(ownerAffiliations: OWNER, isFork: false, first: 100) {
      nodes {
        name
        languages(first: 10, orderBy: {field: SIZE, direction: DESC}) {
          edges {
            size
            node {
              color
              name
            }
          }
        }
      }
    }
  }
}
```

github limit the `first` to be 100, but we can add a parameter `start` to use the pagination and traverse all pages.

For example, we can add edges { cursor } and get a list of cursors

```graphql
{
  user(login: "tc-imba") {
    repositories(ownerAffiliations: OWNER, isFork: false, first: 100) {
      edges {
        cursor
      }
      nodes {
        name
        languages(first: 10, orderBy: {field: SIZE, direction: DESC}) {
          edges {
            size
            node {
              color
              name
            }
          }
        }
      }
    }
  }
}
```

It shows that the cursor of the last repo is `Y3Vyc29yOnYyOpHOCWwxJg==`, and then we can use `start: "Y3Vyc29yOnYyOpHOCWwxJg=="` to find the 101-200 repos, and so on

**Additional context**
I'll draft a PR for this later

Issue #1471, opened 2021-11-24
[Down Time] Card throws 'maximum retries exceeded' error


anuraghazra:
Hi everyone, I see that all personal access tokens are currently failing, thus you will see the "Maximum retries exceeded" error. 

I'm looking into the issue to figure out what caused it. 

Ideally, it should be fixed in an hour when the PATs get rest. 

# Details

## Public instance information

<img alt="Custom badge" src="https://img.shields.io/endpoint?url=https%3A%2F%2Fgithub-readme-stats-git-monitoring-github-readme-stats-team.vercel.app%2Fapi%2Fstatus%2Fup%3Ftype%3Dshields">

## Explanation

-------

Hey everyone, sorry for the recent downtimes. 

For the sake of transparency, we get a huge amount of requests per month (it's kind of mind boggling for a side project which I built just for fun). Here's the vercel dashboard statistics: 

![image](https://user-images.githubusercontent.com/35374649/194407743-9af65825-483a-49af-9e16-cac3a5ea5a50.png)

As you can see even with 73% of the responses being cached by vercel the amount of `live` requests are huge, and the bottleneck here is [Personal Access Tokens](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) 

Each PAT have 5000 points, and based on the complexity of the GraphQL query & [resource limitations](https://docs.github.com/en/graphql/overview/resource-limitations) set by GitHub, it's actually very very low considering that amount of requests. 

To circumvent the issue, we currently have 7 PATs (thats 35k points) which sometimes get exhausted. 
Thanks to @rickstaa we are going to add few more probably bumping it up to 10 or 12, hopefully that could help alleviate the downtimes. 

> Note: That these downtimes are temporary, and only lasts for 1hour, since PATs regenerate after 1 hour

------

*So whats the idea way?*

https://github.com/anuraghazra/github-readme-stats/issues/1471#issuecomment-979306704
Deploy on your own vercel instance. WHY? 
It's easy to do, free & reliable. 

- It reduces loads on our vercel instance.
- Your own PAT will give you more concrete stats including private stats
- It ensures that there won't be any downtime on your profile stats because you are the only one who is gonna hit your URL

_Originally posted by @anuraghazra in https://github.com/anuraghazra/github-readme-stats/issues/2130#issuecomment-1270624232_

jincheng9:
[...]

PR #2111, opened 2022-10-04
fix: Fetch all repos for top languages

MatteoPierro:
This fixed https://github.com/anuraghazra/github-readme-stats/issues/1177

### DIFF
--- a/tests/top-langs.test.js
+++ b/tests/top-langs.test.js
@@ -36,6 +36,10 @@ const data_langs = {
             },
           },
         ],
+        pageInfo: {
+          hasNextPage: false,
+          cursor: "cursor",
+        },
       },
     },
   },
--- a/src/fetchers/top-languages-fetcher.js
+++ b/src/fetchers/top-languages-fetcher.js
@@ -13,10 +13,16 @@ const fetcher = (variables, token) => {
   return request(
     {
       query: `
-      query userInfo($login: String!) {
+      query userInfo($login: String!, $first: Int!, $after: String) {
         user(login: $login) {
           # fetch only owner repos & not forks
-          repositories(ownerAffiliations: OWNER, isFork: false, first: 100) {
+          repositories(
+            ownerAffiliations: OWNER, isFork: false, 
+            first: $first, after: $after
+          ) {
+            edges {
+              cursor
+            }
             nodes {
               name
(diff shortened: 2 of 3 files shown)


MatteoPierro:
Could you review it? @rickstaa 🙏

tests/top-langs.test.js
```diff
@@ -36,6 +36,10 @@ const data_langs = {
             },
           },
         ],
+        pageInfo: {
```
rickstaa, line 39: Maybe let's add tests to check whether the function successfully fetches two pages and whether it stops when the `hasNextPage` parameter becomes `false`?

rickstaa (changes requested):

tests/top-langs.test.js
MatteoPierro: I added a test for fetching two pages

rickstaa: @MatteoPierro Looks good to me. I attached the `hacktoberfest-accepted` label so you should see it show up in your Hacktoberfest profile. We will probably merge this later by introducing an environment variable allowing us to disable this in the public instance (see #2159).

rickstaa (approved):

tc-imba:
LGTM

rickstaa:
@MatteoPierro Thanks for your pull request! I accepted your PR and added the `hacktoberfest-accepted` label. The contribution should appear in your Hacktoberfest dashboard and be accepted in 7 days. 🎉 I will, however, not merge it now due to the PATs rate limit problems we are experiencing (see https://github.com/anuraghazra/github-readme-stats/issues/1471#issuecomment-1271551425). Including multiple pages would deplete the PATs even faster. We might merge it later by adding an environment variable that can be used to disable this feature on the public instance (see https://github.com/anuraghazra/github-readme-stats/pull/2159). For this, we, however, have to run several tests to ensure that the `maxDuration` limit of Vercel is not being hit (see https://vercel.com/blog/customizing-serverless-functions).

mirsella:
> adding an environment variable that can be used to disable this feature

hello ! imo this is the best solution, or an opt-in approach where it's 100 by default, and we can increment the count on our self-hosted instances.
For me, this is the only use case of self-hosting it, and would be useful.

thanks !

tc-imba:
Any updates on this?

PR #2124, opened 2022-10-04
feat: improve card loading speed

rickstaa:
This commit ensures that both the private browser and shared CND caches are used. By doing this, the load on both the Vercel instance, the GraphQL and WakaAPI is reduced. In addition, the `stale-while-revalidate` option was added to speed up the card loading. I now use the vercel recommendations while also keeping a browser cache to lower the demand on the vercel instance (see 
https://vercel.com/docs/concepts/functions/serverless-functions/edge-caching). I am however still researching whether there is a difference between `public, maxage` and `s-maxage` and how `stale-while-revalidate` behaves when no value is supplied.


rickstaa:
@anuraghazra I did some tests and determined that the `stale-while-revalidate` flag requires a value for it to work (compare https://s-w-r-demo.glitch.me/ with https://s-w-r-demo-default.glitch.me/). Further, I found [this issue](https://github.com/symfony/symfony-docs/issues/12934), which explains why it is better to use `public, max-age over `s-maxage`. I, therefore, suggest we only add the `stale-while-revalidate=<24 hours>`. This will not improve our coaching behaviour but increase card loading speeds if multiple requests are made within 24 hours. With this behaviour, people will see the stale server response, but this response is updated in the background for the next visitor.

rickstaa:
@anuraghazra I think the current header makes sense:

```
"Cache-Control: max-age=${cacheSeconds / 2}, s-maxage=${cacheSeconds}, stale-while-revalidate=${CONSTANTS.ONE_DAY}"
```

Using this header, since `s-maxage` is used, the cache is stored both by the browser and the Vercel CDN with different time refresh rates. If the cache is no longer fresh on the browser, it will be requested from the CDN. If the cache is no longer fresh on the CDN, it will return a stale response and update on the server. Using this header, we get the best latency and reduce the load on the GraphQL API. See https://stackoverflow.com/a/74098171/8135687 for more information.

I now have set the private age to be half the CDN time, meaning that the browser is updated two times as frequent as the CDN. Doing this will give users an update faster if they request the cards near the end of the CDN staleness period. However, having the `maxage` and `s-maxage` equal is also an option.

PR #2146, opened 2022-10-06
Add format script to pre-commit hook

HwangTaehyun:
Currently, pre-commit process only includes test process, so when test completed, but remote workflow failed sometimes due to formatting,

Could you review it? @rickstaa 

Thanks!


rickstaa (approved):

rickstaa:
@HwangTaehyun I replaced your configuration with lint-staged so that it only runs on the staged files (see https://github.com/okonet/lint-staged). But apart from that, I think now its good to be merged. Thanks!

PR #2153, opened 2022-10-07
Add types for common directory

HwangTaehyun:
* Add types for common directory and ambient module for emoji-name-map.
Could you review it? @rickstaa @Guri-X @lostgirljourney


anuraghazra:
This will conflict with #2108. 

@rickstaa @HwangTaehyun I would recommend working on stacked PRs to avoid double work and conflicts.

rickstaa:
> This will conflict with #2108.
> 
> @rickstaa @HwangTaehyun I would recommend working on stacked PRs to avoid double work and conflicts.

I was planning to deal with these conflicts myself, but [Stacked pull requests](https://blog.logrocket.com/using-stacked-pull-requests-in-github/) are indeed the superior option :clap:. Thanks for bringing them to my attention! 🚀

tsconfig.json
```diff
@@ -27,11 +27,11 @@
...
     "module": "commonjs",                                /* Specify what module code is generated. */
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    // "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
-    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
+    "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
+    "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
-    // "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
+    "typeRoots": ["./node_modules/@types", "./src/types"],
```
rickstaa, line 34: I always name type folders and the ones in the project folder `@types`. This, however, is just a matter of preference. Let's see what @anuraghazra preferes.

tsconfig.json
```diff
@@ -27,11 +27,11 @@
     /* Modules */
     "module": "commonjs",                                /* Specify what module code is generated. */
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    // "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
-    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
+    "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
+    "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
```
rickstaa, line 31: This also is a matter of preference. Let's leave this decision to @anuraghazra.

tsconfig.json
HwangTaehyun: Sure!

src/cards/types.d.ts
```diff
@@ -49,3 +49,5 @@ type WakaTimeOptions = CommonOptions & {
   layout: "compact" | "normal";
   langs_count: number;
 };
+
+declare module "emoji-name-map";
```
rickstaa, line 53: Maybe we can use a global.d.ts file to add types for modules that don't export types (see https://github.com/rickstaa/github-emoji-picker/blob/main/src/%40types/global.d.ts).

rickstaa: In that case you don't need to add the type folder to each subfolder (i.e. [src/types/emoji-name-map.d.ts](https://github.com/anuraghazra/github-readme-stats/pull/2153/files#diff-de7818750ffd9d577618cebc1f56f789acee279645c6693f698919d8025ffc2b)).

HwangTaehyun: I just have deleted them.

themes/index.ts
```diff
@@ -366,7 +424,8 @@ export const themes = {
     icon_color: "ebbcba",
     text_color: "e0def4",
     bg_color: "191724",
+    border_color: "",
   },
 };
 
-export default themes;
```
rickstaa, line 372: I think we should create a Theme interface and a Themes interface with [an index signature](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures).

src/cards/types.d.ts
HwangTaehyun: It's a good idea to use the global.d.ts like that

src/fetchers/repo-fetcher.ts
```diff
@@ -1,12 +1,13 @@
 // @ts-check
+import { AxiosRequestHeaders } from "axios";
 import { retryer } from "../common/retryer";
 import { MissingParamError, request } from "../common/utils";
 
 /**
- * @param {import('Axios').AxiosRequestHeaders} variables
+ * @param {AxiosRequestHeaders} variables
```
rickstaa, line 7: Since we are now using TS, I was under the impression that we don't use the JSDocs syntax to specify the types in the function description (see https://tsdoc.org).

rickstaa (changes requested):

rickstaa:
I quickly reviewed your PR and requested some changes. I will do a more thorough inspection of each of the types later. I will first wait for @anuraghazra on the stacked PRs.

src/fetchers/repo-fetcher.ts
HwangTaehyun: Yeah, then.. Do you think that we should change all JSDocs to TSDocs?

rickstaa: I think since we don't need the JSDocs syntax anymore, we can remove it. But maybe it is better to do this in a separate PR. Let's see what @anuraghazra suggests.

HwangTaehyun: Ok!

rickstaa (approved):

PR #2160, opened 2022-10-08
fix: prevent errors from being cached

rickstaa:
This PR makes sure that errors are not cached. Doing this will prevent people from seeing the `max_retries` error longer than is needed.


anuraghazra:
good catch. Let's merge this!

Issue #2026, opened 2022-09-15
Hide border option influences commit count


klieret:
## Description

Hide border option influences total  commit count

## Live

![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&hide_border=true&show_icons=true&count_private=true&include_all_commits=true)

![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&hide_border=false&show_icons=true&count_private=true&include_all_commits=true)

![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&show_icons=true&count_private=true&include_all_commits=true)

## Screenshot of the above (15 Sep 2022)

![image](https://user-images.githubusercontent.com/13602468/190458730-fe47a7ae-ea75-4d05-8bed-29e6cd9f55ee.png)

rickstaa:
@klieret Thanks for your bug report. I can not think of how the `hide_border` option can break the card:

https://github.com/anuraghazra/github-readme-stats/blob/3b0f1b11a01e32c4f254861f66a28c14df70c890/src/Card.js#L121

I have heard reports from others (i.e. #1963 and #1968), but those were related to caching or syntax errors. Can you maybe provide me with the syntax you're using for generating the card?

klieret:
Thank you for your quick reply! I'm also quite puzzled. My only hypothesis that there might be some parsing error, perhaps due to wrong syntax on my side that causes some other options to be triggered...

The syntax should be included above (it's live markdown), but let me quickly copy it:

```
![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&hide_border=true&show_icons=true&count_private=true&include_all_commits=true)

![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&hide_border=false&show_icons=true&count_private=true&include_all_commits=true)

![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&show_icons=true&count_private=true&include_all_commits=true)
```

are the three cards.

rickstaa:
@klieret Ah, sorry, I thought all were screenshots. I checked your syntax and it is correct 👍🏻.

**Hide border**

```md
![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=klieret&hide_border=true&show_icons=true&count_private=true&include_all_commits=true)
```

[...]

PR #2177, opened 2022-10-10
fix: prevent totalCommitsFetch error result from being cached

rickstaa:
This PR ensures that when the `https://api.github.com/search/commits?q=author:anuraghazra` API fails, the result is not cached. This prevents people from creating an issue because their commits are 0 for the 4-hour cache period. After this PR is merged, the `0` total commits should only show up when the error occurs.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -50,12 +50,13 @@ export default async (req, res) => {
   }
 
   try {
-    const stats = await fetchStats(
+    const statsResp = await fetchStats(
       username,
       parseBoolean(count_private),
       parseBoolean(include_all_commits),
       parseArray(exclude_repo),
     );
+    const stats = statsResp.stats;
 
     const cacheSeconds = clampValue(
       parseInt(cache_seconds || CONSTANTS.FOUR_HOURS, 10),
@@ -63,7 +64,11 @@ export default async (req, res) => {
       CONSTANTS.ONE_DAY,
     );
 
-    res.setHeader("Cache-Control", `public, max-age=${cacheSeconds}`);
+    if (statsResp.success) {
+      res.setHeader("Cache-Control", `public, max-age=${cacheSeconds}`);
+    } else {
+      res.setHeader("Cache-Control", `no-cache, no-store, must-revalidate`); // Don't cache unsuccessful responses.
+    }
 
     return res.send(
       renderStatsCard(stats, {
--- a/src/fetchers/stats-fetcher.js
+++ b/src/fetchers/stats-fetcher.js
@@ -67,7 +67,7 @@ const fetcher = (variables, token) => {
  *
  * @param {import('axios').AxiosRequestHeaders} variables Fetcher variables.
  * @param {string} token Github token.
- * @returns {Promise<import('../common/types').StatsFetcherResponse>} Repositories fetcher response.
+ * @returns {Promise<import('../common/types').RepositoriesFetcherResponse>} Repositories fetcher response.
  */
 const repositoriesFetcher = (variables, token) => {
   return request(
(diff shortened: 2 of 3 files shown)


rickstaa:
@anuraghazra, maybe we can merge this this will stop people making issues about incorrect total commit count (See https://github.com/anuraghazra/github-readme-stats/issues/2026#issuecomment-1376309417).

api/index.js
```diff
@@ -48,25 +48,32 @@ export default async (req, res) => {
...
+      res.setHeader(
+        "Cache-Control",
+        `max-age=${
+          cacheSeconds / 2
+        }, s-maxage=${cacheSeconds}, stale-while-revalidate=${
+          CONSTANTS.ONE_DAY
+        }`,
+      );
+    } else {
+      res.setHeader("Cache-Control", `no-cache, no-store, must-revalidate`); // Don't cache unsuccessful responses.
```
anuraghazra, line 75: Good idea, but I'm just worried that this will make it so that vercel will need to execute more fresh invocations causing increased load when PATs get exhausted. 

For example say we get 10k fresh requests per day, and we cache 70% of the requests, if PATs get exhausted we will render & cache the error card and after 4hour everything will go back to normal. 

But if we do this, all those 10k requests need to be executed by vercel even though they will all respond with error card. 


I think ideal solution would be to decrease the cache seconds on errors: instead of no-cache let's do 1h for error responses (1h since cache will invalidate the same time as PATs get revalidate)

rickstaa: That is 100% true I overlooked that fact. Let's close this for now.

rickstaa: @anuraghazra In that case we should also change the following code back so that it caches the error:

https://github.com/anuraghazra/github-readme-stats/blob/0ff426d258387ecd0ebbb9206d91b96781e61cf0/api/wakatime.js#L84

https://github.com/anuraghazra/github-readme-stats/blob/0ff426d258387ecd0ebbb9206d91b96781e61cf0/api/top-langs.js#L81

https://github.com/anuraghazra/github-readme-stats/blob/0ff426d258387ecd0ebbb9206d91b96781e61cf0/api/pin.js#L83

https://github.com/anuraghazra/github-readme-stats/blob/0ff426d258387ecd0ebbb9206d91b96781e61cf0/api/index.js#L96

These were merged some time ago and might also be depleting the PATs. However, I suspect this catch block is only triggered when something goes wrong in our code.

rickstaa:
@anuraghazra Closing this for now. See https://github.com/anuraghazra/github-readme-stats/pull/2177#discussion_r1071112093 as it might also be important.

rickstaa:
Re-opening this since we can use this code on private Vercel instances.

api/index.js
Zo-Bro-23: The fetcher handles errors using `throw` as well as `trycatch`, so that means that errors with the API as well as with our code should trigger the `trycatch` in `./api/index.js`, right?

Trying it out in the browser with a random (nonexistent) username gives a no-cache header, so you're right; this should be changed back, since it has the same effect as whatever the PRs trying to do.

rickstaa: @Zo-Bro-23 I changed the code. Please review https://github.com/anuraghazra/github-readme-stats/pull/2448. 👍

Zo-Bro-23: Can we close this PR now since #2448 does the same thing?

rickstaa: This pull request handles errors from the REST API, which will write 0 commits to the card when it fails. While the other pull request handles other errors. I will update this one when #2448 is merged.

rickstaa:
Closing this as I think we should switch to throwing an error when the REST API fails and then use `stale-if-error` to handle the PAT depletion (see https://github.com/anuraghazra/github-readme-stats/pull/2448#issuecomment-1399296739).

Zo-Bro-23:
> Closing this as I think we should switch to throwing an error when the REST API fails and then use `stale-if-error` to handle the PAT depletion (see [#2448 (comment)](https://github.com/anuraghazra/github-readme-stats/pull/2448#issuecomment-1399296739)).

`stale-if-error` seems like a great idea! That will stop the PAT error from showing up during downtime, and since downtime is always less than 1 hour, it can automatically refresh the cache after 1 hour.

api/index.js
Zo-Bro-23: As far as I understand, API errors will trigger Javascript errors, which means that what the other PR does will apply to API errors too.

https://github.com/anuraghazra/github-readme-stats/blob/4b173001264e9774a2179915519e6129af689625/src/fetchers/stats-fetcher.js#L215-L218

The fetcher is throwing a Javascript error, so GraphQL errors should trigger the `trycatch` block too right? Or am I missing something?

PR #2188, opened 2022-10-13
improve folder structure

rickstaa:
Several improvements to the project folder structure:
- refactor: improve folder structure
- docs: rename readme.md to README.md


api/top-langs.js
```diff
@@ -9,7 +9,7 @@ import {
   renderError,
 } from "../src/common/utils.js";
 import { fetchTopLanguages } from "../src/fetchers/top-languages-fetcher.js";
-import { isLocaleAvailable } from "../src/translations.js";
+import { isLocaleAvailable } from "../src/assets/translations.js";
```
anuraghazra, line 12: translations are not an asset though. Would've put it on assets if we had .json locale files but for this it also exports utilities and the locale so let's keep it on the root itself.

tests/api.test.js
```diff
@@ -2,7 +2,7 @@ import { jest } from "@jest/globals";
 import axios from "axios";
 import MockAdapter from "axios-mock-adapter";
 import api from "../api/index.js";
-import { calculateRank } from "../src/calculateRank.js";
+import { calculateRank } from "../src/fetchers/helpers/calculateRank.js";
```
anuraghazra, line 5: maybe calculate rank can be inside common folder, but adding another helper folder just for calculateRank file seems a bit overkill.

rickstaa: Agreed I changed the structure. 👍

api/top-langs.js
rickstaa: Done 👍.

rickstaa:
@anuraghazra I just rebased this and applied your changes.

PR #2241, opened 2022-10-29
Adding Dark_1

Nishanth-thiyakarajan:
Dark_1 has the bg_color as same as the github's dark theme color to match


anuraghazra:
We are currently pausing addition of new themes. If this theme is exclusively for your personal use, then instead of adding it to our theme collection, you can use card [customization options](https://github.com/anuraghazra/github-readme-stats#customization).

PR #2275, opened 2022-11-21
feat: add INCLUDE_ORGS env variable

rickstaa:
This commit adds an INCLUDE_ORGS ENV variable that can be used to include results from organizations into the GRS cards for self-deployed Vercel instances. 

> **Warning**
> This is a hidden feature since we can not officially support this on the Public Vercel instance due to GraphQL and Vercel rate/timeout limits. You can set this env variable via Vercel's ENV variable menu (see https://vercel.com/docs/concepts/projects/environment-variables).


rickstaa:
Nvm, I changed my mind.

Issue #1852, opened 2022-07-07
Stats and languages results incorrect due to graphql 100 records limit


rickstaa:
**Describe the bug**

The current language and stats card results are incorrect for users with more than 100 repositories (see https://docs.github.com/en/graphql/overview/resource-limitations). This is because we are currently only fetching the first 100 repositories of a user account. 

https://github.com/anuraghazra/github-readme-stats/blob/a68abbb9b3a47ad73988f37fce274e882e86b7ad/src/fetchers/stats-fetcher.js#L47

https://github.com/anuraghazra/github-readme-stats/blob/a68abbb9b3a47ad73988f37fce274e882e86b7ad/src/fetchers/top-languages-fetcher.js#L17

Because of this not all languages and stars of a user are represented.

**Expected behaviour**

I think we can use pagination options (i.e. `cursor`, `after`, `endCursor` objects) to loop through all repositories of a user account (see https://stackoverflow.com/a/70039506/8135687). @anuraghazra Let me know what you think.

**TODOs**

- [ ] Allow fetching of multiple pages for the languages data. #2111.
- [x] Improve the stats card such that it fetches more pages when a user has more than 100 repositories with stars. #2100
    - [x]  Merge #2159 to make this available on private Vercel instances while giving us the control to disable it on the public instance because of rate-limiting problems.

rickstaa:
Looks like there already is a PR open for this #1178.

Luminous-Coder:
> For the language card, a PR is already open to solve this #1178. For the stats card, there not yet exists a pull request. This is, however, less important since only people with more than 100 repositories with stars are affected.

Why it is less important? Although people with more than 100 repositories with stars are less.

rickstaa:
@Luminous-Coder Thanks for your question. What I meant is that it is probably affecting fewer people. Since the GraphQL call we make returns the repositories listed from most starred to least starred, a user must have 100 repositories with many stars for to score to be significantly affected. I hope that clears things up. Nevertheless, we should solve both cases. 👍🏻

MatteoPierro:
Hi @rickstaa, is it ok if I work on this for the stats card?

rickstaa:
[...]

PR #2277, opened 2022-11-21
feat: add `include_orgs` card variable

rickstaa:
This pull request adds experimental Organization support. Users can include stats from Organization repositories to which they are collaborators using the `include_orgs` card argument. Please be aware that because of #1852, only the first 100 repositories are used. Including organization, stats might therefore skew your results.

#### TODOs

- [ ] Add a test that checks whether Axios receives the right request query. I tried to google how this is done but could not yet find a good solution. @anuraghazra, do you know how we can implement such a test?

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -21,6 +21,7 @@ export default async (req, res) => {
     show_icons,
     count_private,
     include_all_commits,
+    include_orgs,
     line_height,
     title_color,
     ring_color,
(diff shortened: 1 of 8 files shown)


rickstaa:
@anuraghazra Are there any problems with merging this PRs other than the results being skewed because of https://github.com/anuraghazra/github-readme-stats/issues/1852?

rickstaa:
@anuraghazra Maybe we should add a note about https://github.com/anuraghazra/github-readme-stats/issues/1852 in the readme when merging this PR.

developStorm (approved):

j-i-l (approved):

PR #2302, opened 2022-11-28
Update

bylickilabs:
Update syntax

Co-Authored-By: ®️Bylickilabs™️ <109308073+bylickilabs@users.noreply.github.com>


rickstaa:
@bylickilabs Thanks for your PR. Unfortunately, I think the old version is correct. These are the theme options for the **Repo Card**. I, however, understand where the confusion comes from, as it shows multiple cards. Feel free to elaborate if you think my conclusion is incorrect.

PR #2320, opened 2022-12-09
Corrected Ukrainian translation

okineadev:
Corrected Ukrainian :ukraine: translation and language code (added `ua` to `uk-ua`)


src/translations.js
```diff
@@ -185,7 +190,8 @@ const statCardLocales = ({ name, apostrophe }) => {
       np: "कुल योगदानहरू",
       el: "Συνεισφέρθηκε σε",
       ru: "Внёс вклад в",
-      "uk-ua": "Вніс внесок у",
+      ua: "Внески за рік",
```
rickstaa, line 193: Just wondering. Why is this used instead of Google Translate's `Внесено`. This field can also show all contributions when include_all_commits` is enabled (see [include_all_commits](https://github.com/anuraghazra/github-readme-stats#stats-card-exclusive-options)?

rickstaa (changes requested):
@BogdanDevUA, thanks for your contributions. I think we can merge this if you answer my comment. 🚀

BogdanDevUA:
Ок :smile:

src/translations.js
rickstaa: @BogdanDevUA, what do you think is the google translation correct or the translation which is already there? Further, do you know why there is a `uk-ua` and a UK language code (see https://github.com/ladjs/i18n-locales)? I think we can stick with `uk` since this will just be the Ukrainian language while `uk-ua` will be the ukraining language spoken in Ukraine (see https://github.com/svenfuchs/rails-i18n/issues/285). What do you think?

src/translations.js
```diff
@@ -220,6 +226,7 @@ const repoCardLocales = {
     np: "टेम्पलेट",
     el: "Πρότυπο",
     ru: "Шаблон",
+    ua: "Шаблон",
```
rickstaa, line 229: I think `ua` is incorrect since it specifies Ukraine, not the Ukrainian language.

rickstaa (changes requested):
Please check my comments 👍🏻.

src/translations.js
BogdanDevUA: But for example, ru or el, doesn't it indicate the country or something?
I'm not going to argue with you, that's all.

okineadev:
Oh, I forgor about this PR from 2022

PR #2322, opened 2022-12-10
added black&white theme

corruptzero-dev:
please, add this. i think this theme will look outstanding whatever theme user uses


anuraghazra:
We are currently pausing addition of new themes. If this theme is exclusively for your personal use, then instead of adding it to our theme collection, you can use card [customization options](https://github.com/anuraghazra/github-readme-stats#customization).

PR #2340, opened 2022-12-16
github_dark_green

sultansse:


sultansse:
added new theme

rickstaa (approved):

Zo-Bro-23:
@rickstaa like this? I don't mind merging it.

rickstaa:
> @rickstaa like this? I don't mind merging it.

It is a nice theme which is very distinct from what we already have. I'm only not sure about the name since this theme does not yet appear as a GitHub theme. @sultansse, can you maybe explain where it was based on?

image

PR #2376, opened 2023-01-10
Add light and dark Catppuccin themes with fix for contrast

cosmojg:
I made the minimum viable change to Catppuccin Latte's title color required to pass the AA contrast ratio test.

I would like to note, however, that the original color values passed the WCAG AA contrast ratio test for "Large Text", and the title appears to be "Large Text" in all the README examples. It might make sense just to bypass the test, rather than incorporating this modification, if you want to prioritize remaining faithful to the original color scheme.

Either way, let me know if you need anything else from me! And sorry for the delay!


salesp07:
Light theme looks really good!

anuraghazra:
Although we are pausing theme additions, since this themes are coming from a [community](https://github.com/catppuccin/catppuccin) dedicated to these beautiful theme palettes we will add this.

PR #2378, opened 2023-01-11
Add ruby theme

mike-koala-bear:


MitkoVtori (approved):
This theme looks kinda cool though

themes/index.js
```diff
@@ -374,6 +374,12 @@ export const themes = {
     border_color: "170F0C",
     bg_color: "170F0C",
   },
+  mike_koala_bear: {
```
MitkoVtori, line 377: Change this to **Ruby**

MitkoVtori (approved):
The theme looks kinda cool. I would suggest only one thing, don't name it with your username. As much as I undesrtand from the PR title, this theme is called ruby theme, so I would suggest to rename the theme into ruby.

PR #2409, opened 2023-01-14
Custom Cards

Zo-Bro-23:
# Custom Cards
wild

This PR adds an option to generate custom cards in addition to the GitHub Stats cards. Right now only stats can be generated, but options will be soon added for lanaguage/wakatime and repo cards!

## API Usage
Main endpoint is `https://{VERCEL_URL}/api/wild`. The following options aren't used:
- `username`
- `exclude_repo`
- `locale`
- `count_private`
- `include_all_commits`

In addition, the following new options are added (through query params):
- `totalStars`
- `totalCommits`
- `totalPRs`
- `totalIssues`
- `contributedTo`
- `level` (for rank level)
- `score` (for rank score)
- `starsTitle`
- `commitsTitle`
- `issuesTitle`
- `PRsTitle`
- `contribsTitle`
- `title` (main title)

All added options are optional (default to normal titles and random values), and all previous options are required as was previously.

You may use the `https://github-readme-stats.zohan.tech` deployment to test this PR, but it will constantly be updated as I make changes (so may have errors at some point).

## Changelog
- `./vercel.json` - Changed `maxDuration`; ignore for now (used for testing), will revert before PR is published.
- `./src/cards/stats-card.js`
  - Restructured `kFormatter` to allow `NaN` values too
  - Replaced i18n titles with generic ones and added title as parameters for the `renderStatsCard()` function
  - Replaced specific logic for commits (line 295 old 311 new) with generic logic that is used for all other lines (didn't understand why this was needed - see #2407)
- Added `./api/wild/index.js` - Adds an API endpoint for custom cards like shown above

## To Do
- [ ] Add other card types like language/wakatime and repo cards
- [ ] Add updates to README
- [ ] Add tests
- [ ] Ensure that changing `renderStatsCard()` doesn't affect anything else (other tests, etc)

### DIFF
--- a/api/wild/index.js
+++ b/api/wild/index.js
@@ -1,11 +1,11 @@
-import { renderStatsCard } from "../src/cards/stats-card.js";
+import { renderStatsCard } from "../../src/cards/stats-card.js";
 import {
   clampValue,
   CONSTANTS,
   parseArray,
   parseBoolean,
   renderError,
-} from "../src/common/utils.js";
+} from "../../src/common/utils.js";
 
 export default async (req, res) => {
   const {
--- a/api/wild/index.js
+++ b/api/wild/index.js
@@ -1,5 +1,4 @@
 import { renderStatsCard } from "../src/cards/stats-card.js";
-import { blacklist } from "../src/common/blacklist.js";
 import {
   clampValue,
   CONSTANTS,
@@ -7,20 +6,15 @@ import {
   parseBoolean,
   renderError,
 } from "../src/common/utils.js";
-import { fetchStats } from "../src/fetchers/stats-fetcher.js";
-import { isLocaleAvailable } from "../src/translations.js";
 
 export default async (req, res) => {
   const {
-    username,
     hide,
     hide_title,
     hide_border,
     card_width,
     hide_rank,
     show_icons,
-    count_private,
-    include_all_commits,
     line_height,
     title_color,
     ring_color,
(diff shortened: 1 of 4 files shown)


Zo-Bro-23:
@anuraghazra @rickstaa thoughts?

rickstaa:
@Zo-Bro-23 Thanks for your pull request. I like the idea. I, however, have some very busy months ahead due to my graduation, so I can not promise I can review this PR in time. Maybe @anuraghazra can take a look at it.

Zo-Bro-23:
@anuraghazra ready for review! If possible, I'd like feedback on three things please:
- Parameter names. If you can think of more intuitive variable or parameter names, please let me know. As for that, I think even the API endpoints can be named better, since `top-langs` doesn't actually list languages anymore!
- API integrations. Can you think of any platforms (besides GitHub) that we can integrate with? Now that we have custom cards, it would be really easy to make stat cards for other platforms like GitHub (I can only think of YouTube at the moment).
- README. I've placed this at the bottom of the README, mostly because I don't want this to be obstructive to the proper documentation near the top. Please let me know where you think this should be placed though, since some people might find this useful (even if they don't scroll all the way down to see it), and others might just find it annoying when they're trying to read the documentation for the actual GitHub Stats. Also, let me know if you have any suggestions for the documentation I've written, since I can always try to make it clearer and better!

Looking forward to reading your reviews; cheers!

api/wild/index.js
```diff
@@ -0,0 +1,75 @@
...
+    theme,
+    disable_animations,
+    border_radius,
+    border_color,
+  } = req.query;
+  res.setHeader("Content-Type", "image/svg+xml");
+
+  try {
+    const stats = {
+      totalStars: req.query.total_stars,
```
rickstaa, line 34: I don't think giving users the ability to construct their own `stats` and `language` cards are wise. I think the idea of having an endpoint where people can create custom cards (i.e. `wild`) is very cool. I would, however, make the stats and language card items reserved and throw an error if people try to use those.

rickstaa (changes requested):
I don't think giving users the ability to construct their stats and language cards is wise. The idea of having an endpoint where people can create custom cards (i.e. wild) is very cool. I would, however, make the stats and language card items reserved and throw an error if people try to use those.

PR #2414, opened 2023-01-15
Update readme.md

raphaelricardo10:
fix: missing "&" in show_icons=true in Showing icons section

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -133,7 +133,7 @@ You can add the count of all your private contributions to the total commits cou
 
 ### Showing icons
 
-To enable icons, you can pass `show_icons=true` in the query param, like so:
+To enable icons, you can pass `&show_icons=true` in the query param, like so:
 
 ```md
 ![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=anuraghazra&show_icons=true)


readme.md
```diff
@@ -133,7 +133,7 @@ You can add the count of all your private contributions to the total commits cou
 
 ### Showing icons
 
-To enable icons, you can pass `show_icons=true` in the query param, like so:
+To enable icons, you can pass `&show_icons=true` in the query param, like so:
```
Zo-Bro-23, line 137: Query params follow the structure `?first_param=something&second_param=123&third_param...`. Only if `show_icons` is not the first param sent should you have `&`. Otherwise, you have to use `?`. Order doesn't matter when sending query params, so if someone decides to have `show_icons` before the other params (or as the only param), then it doesn't need the `&`. Because of this ambiguity, documentation usually doesn't include `?` and `&` and only mentions the param. Please feel free to reply if you have any questions.

Zo-Bro-23: I just realized that all the other params in the README have been written with a `&`, so unless you're willing to change all the params, adding a `&` to `show_icons` would probably be the way to go. My bad, sorry!

Zo-Bro-23 (approved):

readme.md
raphaelricardo10: > I just realized that all the other params in the README have been written with a `&`, so unless you're willing to change all the params, adding a `&` to `show_icons` would probably be the way to go. My bad, sorry!

Yes! The first param probably would be the username, so I think it is ok to include the "&" in every optional param.

Zo-Bro-23 (approved):
This is a _really_ small PR, so @rickstaa I think I'll just merge it by myself.

rickstaa:
> This is a _really_ small PR, so @rickstaa I think I'll just merge it by myself.

Excellent call.

Issue #1515, opened 2022-01-02
`Include_all_commits` sometimes shows incorrect number of commits


nasw1h:
**Describe the bug**

On GitHub Stats the number of commits is currently showing the last years commits, after adding `include_all_commits` it decreases as shown below.

**Expected behaviour**

Shows the number of commits in 2022

**Screenshots / Live demo link (paste the github-readme-stats link as markdown image)**

Before adding `include_all_commits`
![image](https://github-readme-stats.vercel.app/api?username=naxwh&theme=dark&show_icons=true&hide_border=true&count_private=true)
After adding `include_all_commits`
![image](https://github-readme-stats.vercel.app/api?username=naxwh&theme=dark&show_icons=true&hide_border=true&count_private=true&include_all_commits=true)

rickstaa:
@nasw1j, I checked the codebase, and I think both `count_private` and `include_all_commits` parameters are correctly implemented.

https://github.com/anuraghazra/github-readme-stats/blob/9a909ff0a41a6abba34b008ce57744d7cb2cd220/src/fetchers/stats-fetcher.js#L155-L163

I, therefore, think the error is caused by incorrect data returned from GitHub's [REST API](https://docs.github.com/en/rest) and [GraphQL API](https://docs.github.com/en/graphql) APIs. If this happens, these wrong values are cached for 4 hours. 

https://github.com/anuraghazra/github-readme-stats/blob/9a909ff0a41a6abba34b008ce57744d7cb2cd220/api/index.js#L60-L64

When this happens, you can check the console.log to see if there is any error and use the `&v=123` syntax to force the cache to reload.

### Detailed explanation

If `include_all_commits` is enabled, and GitHub [REST API](https://docs.github.com/en/rest) fails the code will return 0. 

https://github.com/anuraghazra/github-readme-stats/blob/9a909ff0a41a6abba34b008ce57744d7cb2cd220/src/fetchers/stats-fetcher.js#L69-L100

This will then be used as the total commits. To this value, the private commits will be correctly added if `count_private` is enabled. Since we used caching, this wrong value will be on the card for 4 hours. 

### Posible solutions

We can use the previous cache when this problem happens.

rickstaa:
I quickly checked #1691 created by [Rongronggg9](https://github.com/Rongronggg9) and think merging to his pull request will significantly improve the commit accuracy. I, however, did not have time to do a thorough review 😅.

[...]

Issue #2358, opened 2022-12-30
Bug with commits


vsoch:
### Describe the bug

There seems to be a bug with commits. My (previously S+) chart is now A++ and it only shows 500 commits?

![image](https://user-images.githubusercontent.com/814322/210107462-9c9cd84e-620f-4b2e-a91a-d71b33ae1d4c.png)

I have over 5K.

### Expected behaviour

I have over 5K commits. I wouldn't expect the rating the drop this quickly, and I wouldn't expect the chart to report only 500 or so.

![image](https://user-images.githubusercontent.com/814322/210107516-7802d4eb-896e-410b-abc4-6769aacd74d8.png)

### Screenshots / Live demo link

https://github.com/vsoch/

and 

https://camo.githubusercontent.com/f044fdd12c5db5f147b99799ff57ebd2cdf2438b7ffa1be653a07bc00eeb0b7c/68747470733a2f2f6769746875622d726561646d652d73746174732e76657263656c2e6170702f6170693f757365726e616d653d76736f63682673686f775f69636f6e733d7472756526636f756e745f707269766174653d7472756526696e636c7564655f616c6c5f636f6d6d6974733d74727565267468656d653d7472616e73706172656e74

bylickilabs:
@vsoch 
Hello, It may take some time before the stats are adjusted. Usually every full hour +/-

bylickilabs:
@vsoch 
This is definitely a mistake, but it can happen! My stats graphs currently show 84.8K Total Stars. That's not right either! I am under 130 stars. I would wait and read the statistics again if necessary. Will do the same tomorrow at the latest if it doesn't update. Regards

vsoch:
okay, must be a bug - will circle back tomorrow. Thanks for the speedy response!

vsoch:
Just wanted to post again - the metrics I think are wrong again today?

![image](https://user-images.githubusercontent.com/814322/223886610-fabd1252-955c-4be6-b41a-6c6417dcae11.png)

rickstaa:
> Just wanted to post again - the metrics I think are wrong again today?
> 
> ![image](https://user-images.githubusercontent.com/814322/223886610-fabd1252-955c-4be6-b41a-6c6417dcae11.png)

Thanks! Now we know what is going on (see https://github.com/anuraghazra/github-readme-stats/issues/1515). This happens when the GitHub rest API fails. In that case, `0` is returned, and the private contributions are displayed:

[...]

PR #2448, opened 2023-01-21
feat: rate limit error chaching

rickstaa:
Rate limit error caching to alleviate PATs.


api/index.js
```diff
@@ -93,7 +93,12 @@ export default async (req, res) => {
       }),
     );
   } catch (err) {
-    res.setHeader("Cache-Control", `no-cache, no-store, must-revalidate`); // Don't cache error responses.
+    res.setHeader(
+      "Cache-Control",
+      `max-age=${CONSTANTS.ERROR_CACHE_SECONDS / 2}, s-maxage=${
+        CONSTANTS.ERROR_CACHE_SECONDS
+      }, stale-while-revalidate=${CONSTANTS.ONE_DAY}`,
```
anuraghazra, line 100: Whats happening here? 

`CONSTANTS.ERROR_CACHE_SECONDS / 2`  in max-age
`CONSTANTS.ERROR_CACHE_SECONDS` in s-maxage
`CONSTANTS.ONE_DAY` in revalidate?

anuraghazra: wait this code is on master lol. Have no idea why is it like this 😆

anuraghazra: Okay this was the change https://github.com/anuraghazra/github-readme-stats/pull/2124

rickstaa: Yea, this was when I would sometimes merge PRs without your review since they had been open for a long time. Sorry for the confusion 😅. Feel free to change it to any caching behavior you like. The reasoning for the current caching code can be found at https://github.com/anuraghazra/github-readme-stats/pull/2124#issuecomment-1282864119.

readme.md
```diff
@@ -259,7 +259,7 @@ You can customize the appearance of your `Stats Card` or `Repo Card` however you
 -   `bg_color` - Card's background color _(hex color)_ **or** a gradient in the form of _angle,start,end_. Default: `fffefe`
 -   `hide_border` - Hides the card's border _(boolean)_. Default: `false`
 -   `theme` - name of the theme, choose from [all available themes](./themes/README.md). Default: `default` theme. 
--   `cache_seconds` - set the cache header manually _(min: 7200, max: 86400)_. Default: `14400 seconds (4 hours)`.
+-   `cache_seconds` - set the cache header manually _(min: 14400, max: 86400)_. Default: `14400 seconds (4 hours)`.
 -   `locale` - set the language in the card _(e.g. cn, de, es, etc.)_. Default: `en`.
```
Zo-Bro-23, line 263: Didn't I change this in #2442?

scripts/push-theme-readme.sh
```diff
@@ -9,6 +9,6 @@ git config --global user.name "GitHub Readme Stats Bot"
 git branch -d $BRANCH_NAME || true
 git checkout -b $BRANCH_NAME
 git add --all
-git commit --message "docs(theme): Auto update theme readme" || exit 0
+git commit --no-verify --message "docs(theme): Auto update theme readme"
 git remote add origin-$BRANCH_NAME https://${PERSONAL_TOKEN}@github.com/${GH_REPO}.git
```
Zo-Bro-23, line 13: As far as I can tell, this is unrelated to the rate limiting feature, so it would be best if you could keep these to separate PRs. I also understand that it might be cumbersome to open PRs for small changes, though, so it's fine either way :)

src/cards/top-languages-card.js
```diff
@@ -39,46 +39,53 @@ const getLongestLang = (arr) =>
...
+        x: 0,
+        y: 25,
+        color,
+        width: progressWidth,
+        progress,
+        progressBarBackgroundColor: "#ddd",
+        delay: staggerDelay + 300,
+      })}
+    </g>
   `;
```
Zo-Bro-23, line 70: Seems to be a lot of changes here (unrelated to the rate limit feature). Could you please explain what this does?

src/common/createProgressNode.js
```diff
@@ -19,20 +20,22 @@ const createProgressNode = ({
...
+      <svg data-testid="lang-progress" width="${progressPercentage}%">
+        <rect
+            height="8"
+            fill="${color}"
+            rx="5" ry="5" x="0" y="0"
+            class="lang-progress"
+            style="animation-delay: ${delay}ms;"
+        />
+      </svg>
     </svg>
```
Zo-Bro-23, line 39: Same here (see [this](https://github.com/anuraghazra/github-readme-stats/pull/2448/files#r1083277829)).

src/common/retryer.js
```diff
@@ -45,8 +45,11 @@ const retryer = async (fetcher, variables, retries = 0) => {
     // prettier-ignore
     // also checking for bad credentials if any tokens gets invalidated
     const isBadCredential = err.response.data && err.response.data.message === "Bad credentials";
+    const isAccountSuspended =
+      err.response.data &&
+      err.response.data.message === "Sorry. Your account was suspended.";
 
-    if (isBadCredential) {
+    if (isBadCredential || isAccountSuspended) {
       logger.log(`PAT_${retries + 1} Failed`);
```
Zo-Bro-23, line 53: Seems good, but again, would be better if we could keep different features/fixes to different PRs. Would be easier to merge in that case.

Zo-Bro-23 (approved):
Seems all good!

See my other comments though; particular [this](https://github.com/anuraghazra/github-readme-stats/pull/2448/files#r1083277829).

I'm approving the changes the related to the error caching, but I'm not really sure what some of the other changes do. If you're happy then feel free to merge 🚀

api/index.js
rickstaa: @anuraghazra, we can also remove these lines and keep the same behavior as for the case when no error is thrown. Both are fine with me.

src/cards/top-languages-card.js
Zo-Bro-23: Well, all clear now 🎉

src/common/retryer.js
rickstaa: @Zo-Bro-23 Sorry, that code was due to the PR not being on the main branch. I now rebased it.

src/cards/top-languages-card.js
rickstaa: These changes are already on the master branch. I rebased the PR.

scripts/push-theme-readme.sh
rickstaa: Same already on the master branch.

Zo-Bro-23 (approved):
I'm good for merge 🚀

@anuraghazra thoughts?

rickstaa:
Maybe let's change this to using `if-stale-error` https://developer.fastly.com/learning/concepts/stale/#:~:text=stale%2Dif%2Derror%20tells%20Fastly,during%20periods%20of%20server%20instability (see image below).

image

We can then throw a Vercel error instead of displaying the `max tries` error on the card and let the browser serve the card from the cache (see image below).

https://developer.fastly.com/learning/concepts/stale/#:~:text=stale%2Dif%2Derror%20tells%20Fastly,during%20periods%20of%20server%20instability

anuraghazra:
Yes, but let's test it on dummy deployment before going to prod with this

rickstaa:
> Yes, but let's test it on dummy deployment before going to prod with this

100% 👍🏻.

rickstaa:
> Yes, but let's test it on dummy deployment before going to prod with this

I currently do not have the time to implement this, but it should be doable if we find a way to throw a Vercel error. We can then catch the `MAX_RETRY` and possibly `REST_API_ERROR` (see https://github.com/anuraghazra/github-readme-stats/pull/2177) in the API endpoints and throw this Vercel error. If we add `if-stale-error` to the header, the cards should be served from the cache if the GraphQL or Rest APIs timeout.

rickstaa:
Not sure if `stale-if-error` is already supported with Vercel (see https://github.com/vercel/vercel/issues/3362).

Zo-Bro-23:
I'm not sure if I understand this correctly, but returning a 400 or a 500 error should trigger `stale-if-error` right? There's nothing special we have to do apart from throwing an API error (but I could be wrong).

rickstaa:
> I'm not sure if I understand this correctly, but returning a 400 or a 500 error should trigger `stale-if-error` right? There's nothing special we have to do apart from throwing an API error (but I could be wrong).

I thought the same but up till now I don't know exactly how Vercel deals with errors. You can see what I already tried in:

https://github.com/anuraghazra/github-readme-stats/blob/c504c9338ed830ad619e008ba9ff76422398687f/api/index.js#L104-L108

I can currently respond with a custom error card and attach a `500` status code or show Vercels own `500` error. I was not able to change Vercels error messages. Could be that Vercel shows the previous cached card when a response with a `500` status is given and only shows the error card after the `if-stale-error` period is over but I could not find documentation on this matter and did not yet test. We could try to add a condition that breaks the `retryer` after a certain time. @anuraghazra what do you think?

rickstaa:
Just putting this here for future reference. Fixing this issue will fix our most mentioned bug https://github.com/anuraghazra/github-readme-stats/issues/1515. 👍

rickstaa:
If we like this new method, we can also improve it the [fetchers/stats-fetcher.js/totalCommitsFetcher](https://github.com/anuraghazra/github-readme-stats/blob/c5063b92b6e260dcc405a0a4cd27552dc103f8f5/src/fetchers/stats-fetcher.js#L139-L170) so that it throws an error instead of returning `0`. This will fix #1515 and #2358.

src/common/utils.js
```diff
@@ -293,11 +293,21 @@ const noop = () => {};
 const logger =
   process.env.NODE_ENV !== "test" ? console : { log: noop, error: noop };
 
+// Cache settings.
+const CARD_CACHE_SECONDS = 14400;
+const ERROR_CACHE_SECONDS = 600;
```
anuraghazra, line 298: Let's update this to 30min? Currently it's 10. And if PATs are exhausted that means Vercel will serve fresh responses every 10 minutes, but PATs will only restore after 30min or 1hour.

Changing it to 30min will reduce the load on vercel.

anuraghazra (approved):
Looks good.

anuraghazra:
>I can currently respond with a custom error card and attach a 500 status code or show Vercels own 500 error. I was not able to change Vercels error messages.

Just to confirm we will always be rendering our own custom error SVG card right? not the vercel's error page. If we render vercel's error page that might cause issues since we will not be serving content-type=image/svg at that point.

rickstaa:
Merging this will fix #1515. @qwerty541 I will check this later this week so we can close #1515.

rickstaa:
> > I can currently respond with a custom error card and attach a 500 status code or show Vercels own 500 error. I was not able to change Vercels error messages.
> 
> Just to confirm we will always be rendering our own custom error SVG card right? not the vercel's error page. If we render vercel's error page that might cause issues since we will not be serving content-type=image/svg at that point.

@anuraghazra Yea, in the end I replaced the following code

https://github.com/anuraghazra/github-readme-stats/blob/c504c9338ed830ad619e008ba9ff76422398687f/api/index.js#L99C24-L109

with

https://github.com/anuraghazra/github-readme-stats/blob/0a5ff9647742a4424ebc6574f45d101e0afacec0/api/index.js#L103-L109

src/common/utils.js
rickstaa: @anuraghazra GitHub does restore the individual PATs after 1 hour. However, the frequency at which PATs are refreshed depends on when the PAT was depleted.

image

As a result, the 10 minutes could catch a refreshed token. However, if you think the load of Vercel should be decreased, I am happy to increase it to `1800`.

rickstaa:
Hey @qwerty541 and @anuraghazra,

In our case, it seems that there's no significant advantage to having user requests occur at a frequency twice that of the server, especially since we don't regularly clear the cache or dynamically alter server data. Based on insights from [this source](https://stackoverflow.com/a/15972973/8135687), it appears that we might be able to simplify our caching headers like this:

```js
Cache-control: max-age=CONSTANTS.ERROR_CACHE_SECONDS, stale-while-revalidate=${CONSTANTS.ONE_DAY}
```


However, it's important to note that the information presented in that StackOverflow answer might be outdated, as it conflicts with more recent [specifications](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2.2.8). Unfortunately, I couldn't find any documentation supporting the behaviour described in the new specifications. 

Various sources on the internet suggest using both the `max-age` and `s-maxage` directives. I'm not an expert on caching headers, so I'm unsure whether specifying both `max-age` and `s-maxage` is necessary in our case. Your insights on this would be greatly appreciated!

rickstaa:
@anuraghazra, @qwerty541, this pull request can be merged if we decide which caching period we want to use for errors and whether we can simplify the caching headers 👍🏻.

api/index.js
anuraghazra: Let's keep the same behaviour as old one. Not really confident with how this header will behave on prod.

src/common/utils.js
anuraghazra: Okay i think we can keep it as is, load on Vercel is not a huge deal right now.

anuraghazra (approved):

Issue #2465, opened 2023-01-25
2e2 tests are failing on forked repo


ruchernchong:
### Describe the bug

As per the screenshot below, the test is failing on a forked repo.

### Expected behaviour

Test should pass just like the repo from upstream.

### Screenshots / Live demo link

![image](https://user-images.githubusercontent.com/10343662/214473968-401a9ce0-a56a-4c31-8edd-6a897294313e.png)

rickstaa:
> ### Describe the bug
> As per the screenshot below, the test is failing on a forked repo.
> 
> ### Expected behaviour
> Test should pass just like the repo from upstream.
> 
> ### Screenshots / Live demo link
> ![image](https://user-images.githubusercontent.com/10343662/214473968-401a9ce0-a56a-4c31-8edd-6a897294313e.png)
> 
> ### Additional context
> _No response_

@ruchernchong don't worry; that action is a double-check that the APIs work when a change is deployed to Vercel. It doesn't matter unless you are doing development and want to merge a change into the upstream. More importantly, in this case, the Wakatime API timed out. This happens from time to time. The problem will likely go away when you re-run the action.

ruchernchong:
> > ### Describe the bug
> 
> > As per the screenshot below, the test is failing on a forked repo.
> 
> > 
> 
> > ### Expected behaviour
> 
> > Test should pass just like the repo from upstream.
> 
> > 
> 
> > ### Screenshots / Live demo link
> 
> > ![image](https://user-images.githubusercontent.com/10343662/214473968-401a9ce0-a56a-4c31-8edd-6a897294313e.png)
> 
> > 
> 
> > ### Additional context
> 
> > _No response_
> 
> 
> 
> @ruchernchong don't worry; that action is a double-check that the APIs work when a change is deployed to Vercel. It only matters if you are doing development and want to merge a change into the upstream. More importantly, in this case, the Wakatime API timed out. This happens from time to time. The problem will likely go away when you re-run the action.

Got it. Thank you!

PR #2466, opened 2023-01-25
ci: prevent certain actions from running on forks

rickstaa:
Prevent some actions from running on forks since they are not needed. I left the theme preview/doc, test and e2e actions enabled, as they might be people in development. We can also disable the e2e action if you think it is unnecessary (see https://github.com/anuraghazra/github-readme-stats/issues/2465).

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -5,6 +5,7 @@ on:
 jobs:
   e2eTests:
     if:
+      github.repository == 'anuraghazra/github-readme-stats' &&
       github.event_name == 'deployment_status' &&
       github.event.deployment_status.state == 'success'
     name: Perform 2e2 tests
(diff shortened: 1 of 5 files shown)


ruchernchong (approved):
Thanks for this changes for forked repository!

rickstaa:
> Thanks for this changes for forked repository!

No problem. It, however, does not yet disable the e2e action. I wanted to know what @anuraghazra thinks before doing that. You can, however, manually disable the action in GitHub's action menu. 👍

anuraghazra:
Let's disable e2e for forks since the forked repos will use the Prod URL of GRS which doesn't make sense to use in forks.

anuraghazra (approved):
Lgtm

rickstaa:
> Let's disable e2e for forks since the forked repos will use the Prod URL of GRS which doesn't make sense to use in forks.

Great done thanks! 👍

ruchernchong:
Thanks @rickstaa and @anuraghazra

PR #2471, opened 2023-01-26
Update index.js

littlebutt:
Add a new theme: neon_genesis_evangelion

### DIFF
--- a/themes/index.js
+++ b/themes/index.js
@@ -378,7 +378,7 @@ export const themes = {
     title_color: "8ce360",
     text_color: "ffffff",
     bg_color: "353846",
-    icon_color: "7173e0",
+    icon_color: "9B9DE9",
   }
 };
 
--- a/themes/index.js
+++ b/themes/index.js
@@ -374,6 +374,12 @@ export const themes = {
     border_color: "170F0C",
     bg_color: "170F0C",
   },
+  neon_genesis_evangelion: {
+    title_color: "8ce360",
+    text_color: "ffffff",
+    bg_color: "353846",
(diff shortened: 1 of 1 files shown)


littlebutt (commented):
ok

Zo-Bro-23:
I actually like how this looks, but the colors seem to stand out a bit too much for my taste. Would it be possible to lighten them down by making them a bit more dull? @rickstaa what do you think?

(This looks really similar to what we currently have though, so I'm not sure about cluttering up the theme-base)

rickstaa:
> I actually like how this looks, but the colors seem to stand out a bit too much for my taste. Would it be possible to lighten them down by making them a bit more dull? @rickstaa what do you think?
> 
> (This looks really similar to what we currently have though, so I'm not sure about cluttering up the theme-base)

I agree with your points. To keep the themes library clean, we currently use the following steps for merging theme PRs:

1. If the theme represents a very popular [vscode theme](https://marketplace.visualstudio.com/search?sortBy=Installs&category=Themes&target=VSCode) -> Merge.
2. If this is not the case -> We check if it is different enough from the themes we already have -> NO: close, YES: keep open.
3. We then once in a while, merge the theme PRs that get enough👍🏻.

Currently, all popular themes are already merged (see https://github.com/anuraghazra/github-readme-stats/pulls?q=is%3Apr+is%3Aopen+sort%3Areactions-%2B1-desc+label%3Athemes). There was a proposal by @jesterb0206 to improve new theme visibility, but I didn't yet have the time to implement it (see https://github.com/anuraghazra/github-readme-stats/discussions/2268).

Issue #2179, opened 2022-10-11
Create GRS action


rickstaa:
### Is your feature request related to a problem? Please describe.

Currently, the only way to create GRS cards is through the Public/Private Vercel instance. This method is user-friendly since people can request their cards from the endpoint. Unfortunately, it, however, comes with several limitations:
- It uses the GraphQL API, which is rate limited (see https://docs.github.com/en/graphql/overview/resource-limitations). As a result, if too many people are using the Public Vercel instance simultaneously, cards can no longer be fetched. 
    - Additionally, we can not improve the accuracy of our cards since fetching more data would result in higher GraphQL query costs. Because of this, the results are less accurate for users with more than 100 repositories.
        - https://github.com/anuraghazra/github-readme-stats/issues/1515
        - https://github.com/anuraghazra/github-readme-stats/issues/1852
    - It also prevents us from implementing top feature requests like:
        - https://github.com/anuraghazra/github-readme-stats/issues/1
- Further, the code must run on the Vercel instance before the results are returned. This causes long loading times for the cards or even a time-out.
  
### Describe the solution you'd like

In my discussion with @4l1fe on https://github.com/anuraghazra/github-readme-stats/discussions/1975, we concluded that the best way to solve these issues is also to give users the ability to generate the GRS cards periodically inside a GitHub action. Doing this would provide us to solve the points above for the case when users use GRS through an action.

### Implementation

We can create such an action by calling the render functions directly and then saving the resulting SVGs to the system. I created the following example to show what I mean:

```js
const fetchStats = require("../src/fetchers/stats-fetcher");
const {
  renderError,
  parseBoolean,
  parseArray,
  clampValue,
  CONSTANTS,
} = require("../src/common/utils");
const renderStatsCard = require("../src/cards/stats-card");
const fs = require("fs");

// Variables (Should be replaced with CLI arguments)
const username = "rickstaa";
const count_private = true;
const count_all_commits = true;
const hide = "";
const show_icons = true;
const hide_title = true;
const hide_border = true;
const hide_rank = true;
const include_all_commits = true;
const line_height = 25;
const title_color = "#000000";
const icon_color = "#000000";
const text_color = "#000000";
const bg_color = "#ffffff";
const custom_title = "Rick Staa";
const border_radius = "1";
const border_color = "#000000";
const theme = "light";
const locale = "en";
const disable_animations = true;

const run = async () => {
  // Retrieve stats
  const stats = await fetchStats(
    username,
[...]

PR #2473, opened 2023-01-26
GRS as an npm module!

Zo-Bro-23:
The path towards #2179; releases GRS as an NPM module.

**Note: Currently the NPM module only supports ESM, meaning that CJS users who try to do `require()` will get an error. We should consider using `babel` to automatically convert ESM to CJS before releasing to NPM.**

Things are currently working on my end, and I've published a test package [@zo-bro-23/github-readme-stats-test](https://www.npmjs.com/package/@zo-bro-23/github-readme-stats-test). This is only a draft though, and let's work on more features (such as automatic releases with GH Actions) and fix issues with the current code.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm test
 npx lint-staged
--- a/api/pin.js
+++ b/api/pin.js
@@ -1,4 +1,4 @@
-import { generateRepoCard } from "../src/generators/repo.js";
+import generateRepoCard from "../src/generators/repo.js";
 import { renderError } from "../src/common/utils.js";
 
 export default async (req, res) => {
--- a/api/pin.js
+++ b/api/pin.js
@@ -1,13 +1,5 @@
-import { renderRepoCard } from "../src/cards/repo-card.js";
-import { blacklist } from "../src/common/blacklist.js";
-import {
-  clampValue,
-  CONSTANTS,
-  parseBoolean,
-  renderError,
-} from "../src/common/utils.js";
-import { fetchRepo } from "../src/fetchers/repo-fetcher.js";
-import { isLocaleAvailable } from "../src/translations.js";
+import { generateRepoCard } from "../src/generators/repo.js";
+import { renderError } from "../src/common/utils.js";
 
 export default async (req, res) => {
   const {
@@ -28,35 +20,12 @@ export default async (req, res) => {
 
   res.setHeader("Content-Type", "image/svg+xml");
 
-  if (blacklist.includes(username)) {
-    return res.send(renderError("Something went wrong"));
-  }
-
-  if (locale && !isLocaleAvailable(locale)) {
-    return res.send(renderError("Something went wrong", "Language not found"));
-  }
-
   try {
-    const repoData = await fetchRepo(username, repo);
-
     let cacheSeconds = clampValue(
       parseInt(cache_seconds || CONSTANTS.FOUR_HOURS, 10),
       CONSTANTS.FOUR_HOURS,
       CONSTANTS.ONE_DAY,
     );
-
-    /*
-      if star count & fork count is over 1k then we are kFormating the text
-      and if both are zero we are not showing the stats
-      so we can just make the cache longer, since there is no need to frequent updates
-    */
-    const stars = repoData.starCount;
-    const forks = repoData.forkCount;
-    const isBothOver1K = stars > 1000 && forks > 1000;
-    const isBothUnder1 = stars < 1 && forks < 1;
-    if (!cache_seconds && (isBothOver1K || isBothUnder1)) {
-      cacheSeconds = CONSTANTS.FOUR_HOURS;
-    }
 
     res.setHeader(
       "Cache-Control",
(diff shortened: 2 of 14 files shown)


.husky/pre-commit
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm test
 npx lint-staged
```
Zo-Bro-23, line 4: Kept showing errors when I tried commiting. @rickstaa any idea why that is?

api/pin.js
```diff
@@ -28,57 +20,36 @@ export default async (req, res) => {
...
-    */
-    const stars = repoData.starCount;
-    const forks = repoData.forkCount;
-    const isBothOver1K = stars > 1000 && forks > 1000;
-    const isBothUnder1 = stars < 1 && forks < 1;
-    if (!cache_seconds && (isBothOver1K || isBothUnder1)) {
-      cacheSeconds = CONSTANTS.FOUR_HOURS;
-    }
-
     res.setHeader(
```
Zo-Bro-23, line 30: This doesn't seem to be very effective as the default cache is anyways `FOUR_HOURS`. Let me know if I should add it back.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "github-readme-stats",
-  "version": "1.0.0",
+  "name": "@zo-bro-23/github-readme-stats-test",
+  "version": "1.0.5",
   "description": "Dynamically generate stats for your GitHub readme",
```
Zo-Bro-23, line 4: This is only temporary. I will change it to the actual package name before merging.

src/common/blacklist.js
```diff
@@ -1,4 +1,9 @@
-const blacklist = ["renovate-bot", "technote-space", "sw-yx"];
+const blacklist = [
+  "renovate-bot",
+  "technote-space",
+  "sw-yx",
+  "blacklistedUser", // For testing purposes
+];
```
Zo-Bro-23, line 6: Is this a problem? Makes testing easier.

src/generators/stats.js
```diff
@@ -0,0 +1,75 @@
...
+    custom_title,
+    locale,
+    disable_animations,
+    border_radius,
+    border_color,
+  } = options;
+
+  if (blacklist.includes(username)) {
+    return renderError("Something went wrong");
+  }
```
Zo-Bro-23, line 36: Maybe we should change this to a `throw` instead of a `return`. That way, users can choose to show the error card or not. Currently, if I use the NPM package, this is what my code will look like:

```js
import express from "express";
import { generateStatsCard } from "@zo-bro-23/github-readme-stats-test";

const app = express();

app.get("/", (req, res) => {
  res.setHeader("Content-Type", "image/svg+xml");
  generateStatsCard({ username: "Zo-Bro-23" }).then((card) => {
    res.send(card);
  });
});

app.listen(8000);
```

The `.then()` gets fired for errors too, and even if I added a `.catch()`, it would never get fired. Separating errors and successful responses might be the best way to move forward.

src/index.js
```diff
@@ -1,3 +1,2 @@
-export * from "./common/index.js";
 export * from "./cards/index.js";
-export { getStyles, getAnimations } from "./getStyles.js";
+export * from "./generators/index.js";
```
Zo-Bro-23, line 2: Any reason why you're exporting these? These should be internal functions, and not exported. Keeping these as is bloats up the NPM package when I use `import *`, since everything from the `CONSTANTS` to the `utils` functions gets loaded in.

.husky/pre-commit
rickstaa: I'm not sure what causes the error. You need to update the tests a bit to let them work with your pull request. You can debug these tests with the https://github.com/jest-community/vscode-jest plugin. Sadly, I don't have time to look at it for the next three months 😅.

src/index.js
rickstaa: I wanted to allow other developers to use these functions when they implement new cards. You are right however if the npm package is setup the right way people can just use the `Card` class. 👍🏻

src/generators/stats.js
rickstaa: It makes sense when people use this as an NPM package. 👍🏻 I think @anuraghazra added it this way to fix some renovate problems. 🤔

src/common/blacklist.js
rickstaa: Should not be a problem.

api/pin.js
rickstaa: > This doesn't seem to be very effective as the default cache is anyways `FOUR_HOURS`. Let me know if I should add it back.

Think this should be `EIGHT_HOURS` now that we increased the default to `FOUR_HOURS`.

.husky/pre-commit
rickstaa: This line should stay since we want the tests to run when people try to commit. 🤔

ofek:
Is this almost ready?

rickstaa:
> The path towards #2179; releases GRS as an NPM module.
> 
> **Note: Currently the NPM module only supports ESM, meaning that CJS users who try to do `require()` will get an error. We should consider using `babel` to automatically convert ESM to CJS before releasing to NPM.**
> 
> Things are currently working on my end, and I've published a test package [@zo-bro-23/github-readme-stats-test](https://www.npmjs.com/package/@zo-bro-23/github-readme-stats-test). This is only a draft though, and let's work on more features (such as automatic releases with GH Actions) and fix issues with the current code.

I think the following things still need to be done:

- [ ] Decide whether we want to support CJS modules. 
    In my opinion we shouldn't support such a legacy feature since there hasn't been an npm package in the past. We can choose only to support ESM modules. @Zo-Bro-23, @anuraghazra, @qwerty541, what do you think?
- [ ] I or any of the other collaborators should review this pull request. Other people can check this pull request to help us speed up the merge 👍🏻.
- [ ] @anuraghazra should create a GRS npm organization and release the package.

@Zo-Bro-23 feel free to add other todos I forgot.

PR #2490, opened 2023-02-01
Update .prettierignore

Warrent-waffer:


rickstaa:
@Warrent-waffer I don't see the benefit of adding a hashtag 🤔.

PR #2506, opened 2023-02-08
Add one_dark

b3dyk:
Colors were got from One Dark Pro Theme in the VSCode extention. Please add it to list in oreder to use it in my profile.

### DIFF
--- a/themes/index.js
+++ b/themes/index.js
@@ -374,6 +374,13 @@ export const themes = {
     border_color: "170F0C",
     bg_color: "170F0C",
   },
+  one_dark: {
+    title_color: "E06C75",
+    text_color: "E5C07B",
+    icon_color: "BC78C0",
+    border_color: "3B4048",
+    bg_color: "23272E",
+  },
 };
 
 export default themes;


Zo-Bro-23 (changes requested):
Looks good. ~Please add the theme to `/themes/README.md`.~ Also, in order to not bloat up the themes directory, please choose between this and #2507 (we can't add both since they're extremely similar). I personally prefer this. @rickstaa what do you think? Maybe we can wait to see which one gets more upvotes before deciding. 

Edit: No need to add the theme to the themes README. That happens through an automated workflow. Please just consider which of the two themes you would like to be added to the themes directory (we can't do both).

rickstaa:
> 



> Looks good. ~Please add the theme to `/themes/README.md`.~ Also, in order to not bloat up the themes directory, please choose between this and #2507 (we can't add both since they're extremely similar). I personally prefer this. @rickstaa what do you think? Maybe we can wait to see which one gets more upvotes before deciding.
> 
> Edit: No need to add the theme to the themes README. That happens through an automated workflow. Please just consider which of the two themes you would like to be added to the themes directory (we can't do both).

Closing this as we already have the OneDark theme:

https://github.com/anuraghazra/github-readme-stats/blob/112000667c01f18fd161f204ae3ee796ec2e3011/themes/index.js#L57-L62

b3dyk:
let it be one_dark_pro then. I will delete this one

Issue #2534, opened 2023-02-22
Language card_width is `50px` more than the assigned value.


MD-REHMAN:
### Describe the bug

I found a really small bug, mostly happening because of the below line.

https://github.com/anuraghazra/github-readme-stats/blob/ba7c2f8b55eac8452e479c8bd38b044d204d0424/src/cards/top-languages-card.js#L320

Or maybe because of the default padding.

### Expected behaviour

The card width should not exceed the provided width.

### Screenshots / Live demo link

This can view on my [Profile's README.md](https://github.com/MD-REHMAN)

### Additional context

Expected solution:
1. At least mention the behaviour in docs (README.md).
2. Update the code, so that the added padding would not affect the assigned card width.

MD-REHMAN:
@anuraghazra I can create PR for the fix. But which solution is ideal according to you?

Zo-Bro-23:
![](https://github-readme-stats.vercel.app/api/top-langs/?username=MD-REHMAN&layout=compact&theme=dark&bg_color=0C1116&title_color=F6882B&card_width=445)

<img src="https://github-readme-stats.vercel.app/api/top-langs/?username=MD-REHMAN&layout=compact&theme=dark&bg_color=0C1116&title_color=F6882B" width="445">

Can reproduce.

Zo-Bro-23:
@rickstaa thoughts? @MD-REHMAN feel free to create a PR for this!

Zo-Bro-23:
> @anuraghazra I can create PR for the fix. But which solution is ideal according to you?

I'm not very familiar with SVG and CSS, so I'm not sure if changing padding will break anything else. I'll let @rickstaa decide.

rickstaa:
> > @anuraghazra I can create PR for the fix. But which solution is ideal according to you?
> 
> I'm not very familiar with SVG and CSS, so I'm not sure if changing padding will break anything else. I'll let @rickstaa decide.

I'm aware of this issue, bit didn't think it was a big deal. I, however, understand that people might be confused by the current behaviour. I'm happy to review a PR if somebody creates one. Should be a simple change making sure that the padding is applied while having the width be the width supplied by the user. 👍🏻

rickstaa:
This was fixed in https://github.com/anuraghazra/github-readme-stats/pull/2536. @MD-REHMAN thanks a lot for the fix!

PR #2536, opened 2023-02-22
fix: for issue #2534

MD-REHMAN:
Updated the code and respective test case as well. And also fixed for minimum width difference as well. Here is the preview 

### Without the Fix
Screenshot 2023-02-22 at 5 12 15 PM

### With the Fix
Screenshot 2023-02-22 at 5 11 53 PM

### Increased Min width to accommodate for reduced width
Screenshot 2023-02-22 at 5 38 23 PM

### DIFF
--- a/src/cards/top-languages-card.js
+++ b/src/cards/top-languages-card.js
@@ -13,7 +13,7 @@ import {
 import { langCardLocales } from "../translations.js";
 
 const DEFAULT_CARD_WIDTH = 300;
-const MIN_CARD_WIDTH = 230;
+const MIN_CARD_WIDTH = 280;
 const DEFAULT_LANGS_COUNT = 5;
 const DEFAULT_LANG_COLOR = "#858585";
 const CARD_PADDING = 25;
@@ -317,7 +317,6 @@ const renderTopLanguages = (topLangs, options = {}) => {
 
   let finalLayout = "";
   if (layout === "compact" || hide_progress == true) {
-    width = width + 50; // padding
     height =
       calculateCompactLayoutHeight(langs.length) + (hide_progress ? -25 : 0);
 
--- a/tests/renderTopLanguages.test.js
+++ b/tests/renderTopLanguages.test.js
@@ -216,7 +216,7 @@ describe("Test renderTopLanguages", () => {
     );
     expect(queryAllByTestId(document.body, "lang-progress")[0]).toHaveAttribute(
       "width",
-      "120",
+      "100",
     );
 
     expect(queryAllByTestId(document.body, "lang-name")[1]).toHaveTextContent(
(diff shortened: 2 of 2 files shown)


src/cards/top-languages-card.js
```diff
@@ -317,7 +317,6 @@ const renderTopLanguages = (topLangs, options = {}) => {
 
   let finalLayout = "";
   if (layout === "compact" || hide_progress == true) {
-    width = width + 50; // padding
     height =
```
Zo-Bro-23, line 320: I'm not very familiar with SVG, so could you please explain this for me? You seem to just delete this line, so what was the point of this line in the first place? Any idea why this was written this way? Thinking about that will help us figure out if changing this will break anything else. Thanks for the PR!

tests/renderTopLanguages.test.js
```diff
@@ -216,23 +216,23 @@ describe("Test renderTopLanguages", () => {
...
     );
 
     expect(queryAllByTestId(document.body, "lang-name")[1]).toHaveTextContent(
       "javascript 40.00%",
     );
     expect(queryAllByTestId(document.body, "lang-progress")[1]).toHaveAttribute(
       "width",
-      "120",
+      "100",
     );
```
Zo-Bro-23, line 228: Why are you reducing this by `20` (and `10` below)? Shouldn't you reduce by `50` since that's what the changed code does?

src/cards/top-languages-card.js
MD-REHMAN: I can't be extremely certain why that line was written like that but the most probable guess would be.
The author wanted the `MIN_CARD_WIDTH` to be `230px` but the UI was breaking (as shown in the 3rd screenshot). Thus adding more `50px` to accommodate for the padding.

Anyhow that line is the root cause of the bug. So, I removed that line and increase the `MIN_CARD_WIDTH` by `50`.  Which will fix the case when the card is too small.

tests/renderTopLanguages.test.js
MD-REHMAN: Let me share the math behind it. So the `DEFAULT_CARD_WIDTH = 300`

### Previously 
card_width = 350 (extra 50px)
language_bar_width = 300
HTML is 40% of 300 means 120.
javascript is 40% of 300 means 120.
CSS is 20% of 300 means 60.

### NOW
card_width = 300 (no extra 50px)
language_bar_width = 250 (because of 25px padding on either side)
HTML is 40% of 250 means 100.
javascript is 40% of 250 means 100.
CSS is 20% of 250 means 50.

Zo-Bro-23: Oh, I misunderstood the code; my bad! Makes sense now.

src/cards/top-languages-card.js
Zo-Bro-23: Got it; thanks for the explanation!

Zo-Bro-23 (approved):
Good for merge :+1:

rickstaa (approved):

PR #2549, opened 2023-02-25
fix theme preview

rickstaa:
Fixes the theme-preview GitHub action.

### DIFF
--- a/scripts/preview-theme.js
+++ b/scripts/preview-theme.js
@@ -298,12 +298,13 @@ const themeNameAlreadyExists = (name) => {
   return themes[name] !== undefined;
 };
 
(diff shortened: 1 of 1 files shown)


rickstaa:
@Zo-Bro-23, a small fix for the theme preview action. We can merge it to try it out.

Zo-Bro-23 (approved):
Merging seems fine. Curious though; how will moving the declaration change anything, and why was the workflow working well so far until it suddenly wasn't?

**Edit: Never mind, think I figured it out. Let me know if this is correct: defining the variable inside the try meant that the workflow will fail each time the catch was called, which happened only when themes were invalid. So the workflow sometimes worked and sometimes didn't.**

PR #2551, opened 2023-02-27
Finished EN to IT translation of Vercel setup

Altaro97:


Zo-Bro-23:
We are currently not accepting translation PRs because we are in the process of moving to an automated translation system. Thanks for understanding!

PR #2489, opened 2023-01-31
feat: setup Crowdin integration

andrii-bodnar:
Hey everyone 👋 

According to the discussion in #2053, I'm suggesting an automated solution for the Readme translation based on integration with Crowdin via GH Actions.

In this PR:
- `.github/workflows/localization.yml` - GH Workflows that trigger by pushes to the `master` branch. This workflow uses the [crowdin/github-action](https://github.com/crowdin/github-action) to synchronize localization files (`readme.md`, `docs/*.md`) between the Crowdin Project and GitHub repo.
- `crowdin.yml` - a configuration file for Crowdin synchronization.

Here is the Crowdin **Demo project** - [github-readme-stats-demo](https://crowdin.com/project/github-readme-stats-demo).

In this demo, the existing translations were uploaded to the Crowdin project by the GH Action, and the rest strings were automatically Pre-Translated using the Crowdin NMT machine translation engine (there is a possibility to connect **any MT engine** for that).

In addition, this workflow can be extended to localize all the project texts.

**Example of PR** created by the Action - [#1](https://github.com/andrii-bodnar/github-readme-stats/pull/1/commits/21492f6103cabeed109ff894e35a2a3d8454205f) (Don't worry about the diffs - it's just updated the translation files to the actual state. Crowdin tries to keep the original file structure as much as possible. The next PRs will include the new translations only).

In this integration, there is still a possibility to suggest translations by the community. These translations will be synchronized from Crowdin to the Repo automatically. By the way, Crowdin provides an easy and comfortable **preview** of files for translators. This makes the translation process extremely straightforward.

image

PR #2575, opened 2023-03-12
docs: Correct language code in readme filename

Luminous-Coder:
Change readme filename from "cn" to "zh-CN" that represents Chinese variant of Chinese; follow the format used for "pt-BR".


Zo-Bro-23:
Why only Chinese though? Do it for all languages then.

Luminous-Coder:
> Why only Chinese though? Do it for all languages then.

I don't know which variants those translations use.

Zo-Bro-23:
> > Why only Chinese though? Do it for all languages then.
> 
> I don't know which variants those translations use.

Fair point. We'll see what we can do.

rickstaa:
> > > Why only Chinese though? Do it for all languages then.
> > 
> > 
> > I don't know which variants those translations use.
> 
> Fair point. We'll see what we can do.

@Luminous-Coder, thanks for your PR. However, the translation files will be deprecated when https://github.com/anuraghazra/github-readme-stats/pull/2489 is merged.

Luminous-Coder:
> 

Got it.

If https://github.com/anuraghazra/github-readme-stats/pull/2489 is merged, people translate on Crowdin and GitHub action will automatically update, right?

Zo-Bro-23:
> > > > > Why only Chinese though? Do it for all languages then.
> > > > 
> > > > 
> > > > I don't know which variants those translations use.
> > > 
> > > 
> > > Fair point. We'll see what we can do.
> > 
> > 
> > @Luminous-Coder, thanks for your PR. However, the translation files will be deprecated when #2489 is merged.
> 
> Got it.
> 
> If #2489 is merged, people translate on Crowdin and GitHub action will automatically update, right?

Yes. Closing this for now.

lumynou5:
Hello, I notice the Crowdin PR ended up being closed, and the current `CONTRIBUTING.md` specifies how to contribute to the translation of the software strings but not the documents.  It also says the language codes should follow ISO 639-1, and `cn` obviously doesn't.  Deprecating it in the API would be a big change, but maybe we can at least correct that in the documents?  If so, please consider to reopen this PR.

---

> Why only Chinese though? Do it for all languages then.

This question has another answer actually: Some languages don't need to have a region code.  They probably have no variants with considerable differences.  For example, `ja-JP` means "Japanese, used in Japan", which is kind of meaningless.

PR #2576, opened 2023-03-13
update my theme

eduardozaniboni:


Zo-Bro-23:
@rickstaa will this have backward compatibility issues?

rickstaa:
> @ricksta,a will this have backward compatibility issues?

We can ignore the theme preview action since it must be coded for theme renames. For backward compatibility, the rename will cause the cards of people using the theme to revert to the default theme.

Zo-Bro-23:
> 

Yes, but this theme was added recently, so it should be fine right?

rickstaa:
> > 
> 
> Yes, but this theme was added recently, so it should be fine right?

Yea no problem. I will approve.

rickstaa (approved):

PR #2612, opened 2023-04-13
Added night_vibe theme

MitkoVtori:


MitkoVtori:
Nope, I want the card that way.

PR #2649, opened 2023-04-26
Fixed todo inside stats card data fetcher

qwerty541:


src/fetchers/stats-fetcher.js
```diff
@@ -259,7 +259,6 @@ const fetchStats = async (
       return prev + curr.stargazers.totalCount;
     }, 0);
 
-  // @ts-ignore // TODO: Fix this.
```
rickstaa, line 262: @qwerty541 Did you check whether this issue is already resolved? I added this because Vscode gave an TS error which I wanted to suppress.

rickstaa (changes requested):

src/fetchers/stats-fetcher.js
qwerty541: @rickstaa If you remove this ts-ignore attribute vscode shows the following error:

```
Argument of type '{ totalCommits: number; totalRepos: any; followers: any; contributions: number; stargazers: number; prs: number; issues: number; }' is not assignable to parameter of type 'number'.ts(2345)
```

This occurs because of incorrect docstring for `calculateRank` function. There defined that this function accept seven arguments, but actually it accept an OBJECT with seven properties. I have fixed docstring for this function and this ts-ignore attribute can be removed.

rickstaa: @qwerty541 Thanks for the explanation. I forgot what error was thrown 👍🏻.

rickstaa (approved):

PR #2661, opened 2023-04-29
Add PRs cache cleaning workflow

qwerty541:
This workflow will delete all caches related to PR when it was merged or closed.

### DIFF
--- a/.github/workflows/prs-cache-clean.yml
+++ b/.github/workflows/prs-cache-clean.yml
@@ -1,0 +1,33 @@
+name: prs cache clean
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+          BRANCH="refs/pull/${{ github.event.pull_request.number }}/merge"
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys. 
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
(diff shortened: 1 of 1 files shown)


rickstaa:
@qwerty541 Just curious why should we manually clean the cache when GitHub automatically does this every 7 days 🤔 (see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#:~:text=the%20main%20branch-,Usage%20limits%20and%20eviction%20policy,is%20limited%20to%2010%20GB)?

qwerty541:
> @qwerty541 Just curious why should we manually clean the cache when GitHub automatically does this every 7 days 🤔 (see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)?

@rickstaa I am aware that GitHub automatically deletes the cache, but I would put the question differently. Why do we need to keep the cache for seven days after the merge or the closing of the pull request if we know for sure that we no longer need it? Especially considering that dependabot automatic security and dependency updates are enabled for the repository and there are a lot of pull requests from users. Potentially, a lot of such junk cache can be accumulate. I agree that having such workflow is optional, but I prefer to use this workflow in my repositories to keep things organized. Close this pull request if you think it's excess.

rickstaa:
> > @qwerty541 Just curious why should we manually clean the cache when GitHub automatically does this every 7 days 🤔 (see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)?
> 
> @rickstaa I am aware that GitHub automatically deletes the cache, but I would put the question differently. Why do we need to keep the cache for seven days after the merge or the closing of the pull request if we know for sure that we no longer need it? Especially considering that dependabot automatic security and dependency updates are enabled for the repository and there are a lot of pull requests from users. Potentially, a lot of such junk cache can be accumulate. I agree that having such workflow is optional, but I prefer to use this workflow in my repositories to keep things organized. Close this pull request if you think it's excess.

I'm okay with merging it. Will review later. I was just wondering what the reason was because as far as I am aware, the extra cache does not give problems on our side. Does less cache lead to quicker runtimes 🤔?

rickstaa (approved):

PR #2670, opened 2023-05-01
fix(wakatime-fetcher): add default range to prevent user not found error

fabianocouto:
# Description

This update:
- add default wakatime range to prevent user not found error
- update demos in readme

# Bug Description

Without range param, it returns user not found error.

https://github-readme-stats.vercel.app/api/wakatime?username=fabianocouto

image

With range param, card works as expected.

https://github-readme-stats.vercel.app/api/wakatime?username=fabianocouto&range=all_time

image

# After fix

The option `all_time` is defined as default range if this is not informed.

https://fabianocouto-readme-stats.vercel.app/api/wakatime?username=fabianocouto

image


readme.md
```diff
@@ -456,13 +456,13 @@ Change the `?username=` value to your [Wakatime](https://wakatime.com) username.
 
 ### Demo
 
-[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=willianrod)](https://github.com/anuraghazra/github-readme-stats)
+[![Harlok's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&range=all_time)](https://github.com/anuraghazra/github-readme-stats)
```
rickstaa, line 459: The pull request looks good to me. However, why are you also adding `&range=all_time` here when it is already fixed in:

https://github.com/anuraghazra/github-readme-stats/blob/0057714068719cc0d12ddf6f5069147eecd2ff80/src/fetchers/wakatime-fetcher.js#L17-L19

rickstaa (changes requested):
Thanks for your pull request. It looks good to me, but I think `range=all_time` does not need to be added to the documentation. After that is removed, I will merge.

MullerHub (approved):

readme.md
MullerHub: It`s not working at me

fabianocouto: > The pull request looks good to me. However, why are you also adding `&range=all_time` here when it is already fixed in:
> 
> https://github.com/anuraghazra/github-readme-stats/blob/0057714068719cc0d12ddf6f5069147eecd2ff80/src/fetchers/wakatime-fetcher.js#L17-L19

You're right! It will not necessary after fix.

fabianocouto:
> Thanks for your pull request. It looks good to me, but I think `range=all_time` does not need to be added to the documentation. After that is removed, I will merge.

It is done! @rickstaa

readme.md
fabianocouto: > It`s not working at me

Hi @MullerHub !
Can you explain the problem in more detail?
Maybe with examples links.

rickstaa (approved):

PR #2681, opened 2023-05-04
fix: added '&range=last_7_days' to fix 'user not found' error

aylmerbolzan:
Even if `&range=last_7_days` is the default, for some reason it stopped working, being necessary to declare it directly in the code.

`&range=last_7_days` has been added to the main readme as well as readmes in the docs folder in other languages.


rickstaa:
Thanks for your pull request! This, however, will already be fixed when #2670 is merged.

Issue #2682, opened 2023-05-04
readme- stats waka card isn't working 


abirett2:
### Describe the bug

waka stats card isn't working

vitorhonna:
Try to add the parameter `&range=all_time` to the url.

Check #2671 and #2668 for further info.

rickstaa:
It should be fixed now that #2662 was merged.

Issue #2671, opened 2023-05-01
wakatime card stopped working suddenly


jNaimXIII:
I've been using the wakatime card for a while to show my most used languages. it randomly started showing an error saying my username doesn't exist even though it's completely fine.

<img width="646" alt="image" src="https://user-images.githubusercontent.com/115424698/235397553-3f16a3af-49a9-44b3-9026-ceaf26a6e8fc.png">
<img width="1199" alt="image" src="https://user-images.githubusercontent.com/115424698/235397591-eb03c0c3-b03a-4878-946d-9ce8a9d953d9.png">

https://wakatime.com/jNaimXIII

brutalzinn:
I having the same issue.

Enaium:
same here

Yoak3n:
same issue

sabin5105:
Me either

brutalzinn:
Wakatime reported issues on generate stats yesterday at 01:30 AM.

chrisK824:
+1 here

forgetfulskybro:
+1 here

ArikSetyawan:
same issue

SwiftlyAside:
same here.

Aleksey-Voko:
Adding a parameter worked for me: `&range=all_time`

aylmerbolzan:
Even putting the `&range=all_time` parameter at the end of the URL didn't work for me 🤔

HassanYu7:
> Even putting the `&range=all_time` parameter at the end of the URL didn't work for me 🤔

Me either

Aleksey-Voko:
> Even putting the `&range=all_time` parameter at the end of the URL didn't work for me thinking

Perhaps this is due to the presence of compound parameters in the URL, where the values are separated by commas.

`&hide=text,git,other,xml&...`

If you remove such parameters, it will work.
Or maybe somehow enclose them in quotation marks...
Or move to the beginning, before commas...

alidehkhodaei:
I having the same issue.

alidehkhodaei:
Hello, i found a solution.

Add the parameter 'range=last_7_days' and langs_count=5 to the URL.
'range=last_year' or  'range=all_time'  also works.

You can see my link.
https://github-readme-stats.vercel.app/api/wakatime?username=alidehkhodaei&theme=dark&langs_count=5&range=last_7_days

aylmerbolzan:
That's weird, @alidehkhodaei.

[...]

PR #2694, opened 2023-05-05
feat: clarify wakatime 403 error

### LINKED ISSUE #2683: Wakatime stats  is not working
### Describe the bug

The card that I use is this: https://github-readme-stats.vercel.app/api/wakatime?username=IsfetShen&theme=midnight-purple 

I checked the readme official and there's also not working

### Expected behaviour

_No response_

image

### Additional context

_No response_

mrtnvgr:
Closes #2683, #2682, #2671

### DIFF
--- a/src/fetchers/wakatime-fetcher.js
+++ b/src/fetchers/wakatime-fetcher.js
@@ -19,7 +19,11 @@ const fetchWakatimeStats = async ({ username, api_domain, range }) => {
 
     return data.data;
   } catch (err) {
-    if (err.response.status < 200 || err.response.status > 299) {
+    if (err.response.status == 403) {
+      throw new Error(
+        'History for this time period is not publicly available. Please ensure that the query parameter matches the value in the user\'s profile or try using the "last_7_days" range parameter.',
+      );
+    } else if (err.response.status < 200 || err.response.status > 299) {
       throw new Error(
         "Wakatime user not found, make sure you have a wakatime profile",
       );


KevinGuancheDarias:
I think the message would be misleading as the last_7_days option can too provide the same error for some users, it looks like the range option stopped been flexible by the wakatime api, as only the configured value in the wakatime seems to work, for example I have all_time in wakatime,  so only all_time will work in *range* request param but not the last_7_days nor last_30_days values

image


I would set the message to something like "The specified range doesn't match the time range configured in your wakatime settings to be publicly available", or if wakatime doesn't fix their side, I would directly remove the range option and use the one configured in the profile

mrtnvgr:
> I think the message would be misleading as the last_7_days option can too provide the same error for some users, it looks like the range option stopped been flexible by the wakatime api, as only the configured value in the wakatime seems to work, for example I have all_time in wakatime,  so only all_time will work in *range* request param but not the last_7_days nor last_30_days values
> 
> image
> 
> 
> I would set the message to something like "The specified range doesn't match the time range configured in your wakatime settings to be publicly available", or if wakatime doesn't fix their side, I would directly remove the range option and use the one configured in the profile
> 
> Edit: Wakatime API seems to work with all last_7_days, last_30_days, and all_time

Thank you. I want to add that 403 is returned when the range is not specified in the API call parameters.

KevinGuancheDarias:
Hello @mrtnvgr I have made a mistake in my "Edit" section, wakatime only works with the configured setting, It was working because I was logged in Wakatime, but after I logged out, i get the same behavior than the github-readme-stats project

# Won't work throwing 403 (as setting is set to all_time)
image
image

# Working
image

# When changing the settings to last 7 days in the wakatime settings
image

And then... 😱 
image

mrtnvgr:
@KevinGuancheDarias I updated the error, can you check it? Is it too long?

rickstaa:
Closing this as this should bed fixed already.

Issue #2698, opened 2023-05-06
Waka time card broken range parameter broken


rickstaa:
### Describe the bug

It looks like the `range` parameter of the WakaTime card is no longer working due to some upstream changes. https://github.com/anuraghazra/github-readme-stats/pull/2670 fixed the card for the `all_time` range. For the other ranges, however, the 403 "Forbidden" error is thrown. The strange thing is that this error is not thrown when I directly try the generated WakaTime API link in the browser, but it only happens when I request it through Axios. This is strange as according to the documentation this should work.

https://github.com/anuraghazra/github-readme-stats/blob/688f4e497593d22d813423248bd96aaf0f0e495f/src/fetchers/wakatime-fetcher.js#L14-L20

**Regular card**

[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact)](https://github.com/anuraghazra/github-readme-stats)

```md
[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact)](https://github.com/anuraghazra/github-readme-stats)
```

***All time**

[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=all_time)](https://github.com/anuraghazra/github-readme-stats)

```md
[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=all_time)](https://github.com/anuraghazra/github-readme-stats)
```

**Last 7 days**

[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=last_7_days)](https://github.com/anuraghazra/github-readme-stats)

```md
[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=last_7_days)](https://github.com/anuraghazra/github-readme-stats)
```

**Last 30 days**

[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=last_30_days)](https://github.com/anuraghazra/github-readme-stats)

```md
[![willianrod's wakatime stats](https://github-readme-stats.vercel.app/api/wakatime?username=Harlok&layout=compact&range=last_30_days)](https://github.com/anuraghazra/github-readme-stats)
```

### Expected behaviour

[...]

PR #2707, opened 2023-05-08
fix: Make WakaTime card compatible with new API

rickstaa:
This commit makes sure that the WakaTime card works with the new WakaTime API. See https://github.com/anuraghazra/github-readme-stats/issues/2698 for more information.

>  **Note**
> Will be merged when I have one review of somebody that does use the WakaTime card.

### DIFF
--- a/src/cards/wakatime-card.js
+++ b/src/cards/wakatime-card.js
@@ -279,7 +279,11 @@ const renderWakatimeCard = (stats = {}, options = { hide: [] }) => {
           : noCodingActivityNode({
               // @ts-ignore
               color: textColor,
-              text: i18n.t("wakatimecard.nocodingactivity"),
+              text: !stats.is_coding_activity_visible
+                ? i18n.t("wakatimecard.notpublic")
+                : stats.is_other_usage_visible
+                ? i18n.t("wakatimecard.nocodingactivity")
+                : i18n.t("wakatimecard.nocodedetails"),
             })
       }
     `;
@@ -304,7 +308,11 @@ const renderWakatimeCard = (stats = {}, options = { hide: [] }) => {
             noCodingActivityNode({
               // @ts-ignore
               color: textColor,
-              text: i18n.t("wakatimecard.nocodingactivity"),
+              text: !stats.is_coding_activity_visible
+                ? i18n.t("wakatimecard.notpublic")
+                : stats.is_other_usage_visible
+                ? i18n.t("wakatimecard.nocodingactivity")
+                : i18n.t("wakatimecard.nocodedetails"),
             }),
           ],
       gap: lheight,
@@ -312,9 +320,20 @@ const renderWakatimeCard = (stats = {}, options = { hide: [] }) => {
     }).join("");
   }
 
+  // Get title range text
+  let titleText = i18n.t("wakatimecard.title");
+  switch (stats.range) {
+    case "last_7_days":
+      titleText += ` (${i18n.t("wakatimecard.last7days")})`;
+      break;
+    case "last_year":
+      titleText += ` (${i18n.t("wakatimecard.lastyear")})`;
+      break;
+  }
+
   const card = new Card({
     customTitle: custom_title,
-    defaultTitle: i18n.t("wakatimecard.title"),
+    defaultTitle: titleText,
     width: 495,
     height,
     border_radius,
--- a/src/fetchers/wakatime-fetcher.js
+++ b/src/fetchers/wakatime-fetcher.js
@@ -4,29 +4,30 @@ import { MissingParamError } from "../common/utils.js";
 /**
  * WakaTime data fetcher.
  *
- * @param {{username: string, api_domain: string, range: string}} props Fetcher props.
+ * @param {{username: string, api_domain: string}} props Fetcher props.
  * @returns {Promise<WakaTimeData>} WakaTime data response.
  */
-const fetchWakatimeStats = async ({ username, api_domain, range }) => {
+const fetchWakatimeStats = async ({ username, api_domain }) => {
   if (!username) throw new MissingParamError(["username"]);
 
-  try {
-    const { data } = await axios.get(
-      `https://${
-        api_domain ? api_domain.replace(/\/$/gi, "") : "wakatime.com"
-      }/api/v1/users/${username}/stats/${
-        range || "all_time"
-      }?is_including_today=true`,
-    );
-
-    return data.data;
-  } catch (err) {
-    if (err.response.status < 200 || err.response.status > 299) {
-      throw new Error(
-        "Wakatime user not found, make sure you have a wakatime profile",
+  // Loop through available ranges to get user data.
+  for (const range of ["all_time", "last_7_days", "last_year"]) {
+    try {
+      const { data } = await axios.get(
+        `https://${
+          api_domain ? api_domain.replace(/\/$/gi, "") : "wakatime.com"
+        }/api/v1/users/${username}/stats/${range}?is_including_today=true`,
       );
+
+      return data.data;
+    } catch (err) {
+      if (err.response.status === 403 && range !== "last_year") {
+        continue;
+      } else if (err.response.status < 200 || err.response.status > 299) {
+        throw new Error(i18n.t("wakatimecard.nouser"));
+      }
+      throw err;
     }
-    throw err;
   }
 };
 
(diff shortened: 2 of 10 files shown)


src/fetchers/wakatime-fetcher.js
```diff
@@ -4,29 +4,30 @@ import { MissingParamError } from "../common/utils.js";
...
-      }?is_including_today=true`,
-    );
-
-    return data.data;
-  } catch (err) {
-    if (err.response.status < 200 || err.response.status > 299) {
-      throw new Error(
-        "Wakatime user not found, make sure you have a wakatime profile",
+  // Loop through available ranges to get user data.
+  for (const range of ["all_time", "last_7_days", "last_year"]) {
```
hakula139, line 14: Perhaps swap the order of "last_year" and "last_7_days", since if the user sets the preference to option "Last year", we may be accessible to both "last_year" and the narrower range "last_7_days", when in fact the user might want to show the range "last_year". Therefore an order from a wider range to a narrower one may sound better.

src/translations.js
```diff
@@ -326,6 +326,26 @@ const wakatimeCardLocales = {
...
+    en: "last year",
+    cn: "去年",
+  },
+  "wakatimecard.last7days": {
+    en: "last 7 days",
+    cn: "最近 7 天",
+  },
+  "wakatimecard.nouser": {
+    en: "Wakatime user not found, make sure you have a Wakatime profile",
+    cn: "未找到 Wakatime 用戶，請確保您擁有 Wakatime 個人資料",
```
hakula139, line 339: This (and the following ones) should be "zh-tw", when "cn" refers to Chinese (Simplified).

hakula139 (changes requested):

src/translations.js
rickstaa:
```suggestion
    cn: "未找到 Wakatime 用户，请确保您拥有 Wakatime 个人资料",
```

rickstaa: Thanks! I don't speak Chinese, so I assumed the short abbreviation was the traditional one.

rickstaa: It should be fixed in [7ea55d0](https://github.com/anuraghazra/github-readme-stats/pull/2707/commits/7ea55d05ade794fe7430c8eb2e50b2f5885d9fdb).

src/fetchers/wakatime-fetcher.js
rickstaa: @hakula139 I performed some testing, and it looks like Waka Time only allows people to access the range that is public. I, however, agree that from a time perspective it makes sense to swap the two 👍🏻.

rickstaa: Fixed in [d80d766](https://github.com/anuraghazra/github-readme-stats/pull/2707/commits/d80d7668b6770ca27b637cebd7ae95da407a94f0).

src/fetchers/wakatime-fetcher.js
```diff
@@ -4,29 +4,30 @@ import { MissingParamError } from "../common/utils.js";
...
-      }?is_including_today=true`,
-    );
-
-    return data.data;
-  } catch (err) {
-    if (err.response.status < 200 || err.response.status > 299) {
-      throw new Error(
-        "Wakatime user not found, make sure you have a wakatime profile",
+  // Loop through available ranges to get user data.
+  for (const range of ["all_time", "last_year", "last_7_days"]) {
```
alanhamlett, line 14: Instead of a loop you can also fetch https://wakatime.com/api/v1/users/current and use `public_profile_time_range` to find the user's public profile range. It might end up being the same performance, sometimes taking one request more and sometimes saving one request depending on how many tries before this loop finds the right range.

rickstaa: @alanhamlett I agree using that endpoint is better. Unfortunately, last time I checked, Wakatime decided to put this endpoint behind an auth wall 😓. When I try to fetch `https://wakatime.com/api/v1/users/rickstaa` from within GRS for example, I get the following error:

```json
{"error": "Unauthorized"}
```

I think the only way to use this API is to add an authentication token.

src/cards/wakatime-card.js
```diff
@@ -304,17 +308,32 @@ const renderWakatimeCard = (stats = {}, options = { hide: [] }) => {
...
             }),
           ],
       gap: lheight,
       direction: "column",
     }).join("");
   }
 
+  // Get title range text
+  let titleText = i18n.t("wakatimecard.title");
+  switch (stats.range) {
```
hakula139, line 325: It seems that the `range` parameter has been removed, so will `stats.range` always be undefined here? Since we did not actually assign it during the loop, and it might not be provided by the user either. If so, maybe we also need to modify the relevant mock data (and their expected responses) in the e2e test, as one of them seems to have failed (and `range` parameter should not be provided in mock data).

hakula139 (changes requested):

src/fetchers/wakatime-fetcher.js
alanhamlett: Oh yea, then your loop solution is the best way.

src/cards/wakatime-card.js
alanhamlett: `range` attribute is still returned in `Stats` endpoint:

https://wakatime.com/developers/#stats

hakula139: > `range` attribute is still returned in `Stats` endpoint:
> 
> https://wakatime.com/developers/#stats

Get it. When debugging the failure in e2e test, I thought `range` was from the query by mistake. Thanks for pointing out!

hakula139: It seems like [catelinemnemosyne](https://wakatime.com/@catelinemnemosyne) is using the `range` option `all_time` now, which causes the e2e test to fail, so simply modify the mock data should be fine.

src/fetchers/wakatime-fetcher.js
rickstaa: Because of your comment, I looked into the API a bit more and found out that we can get the range from the https://wakatime.com/api/v1/users/rickstaa/stats. I created 6b8e06b2e40d74be58003cbb6897729a7fa1010d to apply your suggestions.

src/cards/wakatime-card.js
rickstaa: Yea, GitHub totally lost it for some minutes 🤣. I think I fixed all your issues in [98fcc65](https://github.com/anuraghazra/github-readme-stats/pull/2707/commits/98fcc651138c0d6ae5cc68c3dbf8cc58849ccf75).

src/fetchers/wakatime-fetcher.js
```diff
@@ -4,27 +4,32 @@ import { MissingParamError } from "../common/utils.js";
...
 
+  // Retrieve user range.
+  const { data: userdata } = await axios.get(
+    `https://${
+      api_domain ? api_domain.replace(/\/$/gi, "") : "wakatime.com"
+    }/api/v1/users/${username}/stats`,
+  );
+  const range = userdata.data.range;
+
+  // Loop through available ranges to get user data.
```
hakula139, line 21: Seems like it has been fixed from the upstream. Do we still need the following request here?

hakula139 (changes requested):

tests/fetchWakatime.test.js
```diff
@@ -103,6 +134,9 @@ const wakaTimeData = {
 describe("Wakatime fetcher", () => {
   it("should fetch correct wakatime data", async () => {
     const username = "anuraghazra";
```
hakula139, line 136: Should also be changed here.

hakula139 (changes requested):

src/fetchers/wakatime-fetcher.js
rickstaa: Good catch, I think the endpoint was overlooked in https://github.com/anuraghazra/github-readme-stats/commit/a340900e8a1fac282d953fd4c7ac1ae9a459726e.

hakula139: `/stats` is fine, but `/stats/` is not 😄

hakula139 (approved):

JiveOff:
Can confirm this MR fixes the issue I've been having with my card under my Wakapi instance.

rickstaa:
> Can confirm this MR fixes the issue I've been having with my card under my Wakapi instance.

@JiveOff thanks for the confirmation!

muety:
Any chances we can add the `range` query param back in? At least from Wakapi's perspective, I'd find it super convenient if a user can customize the time range displayed by their readme card _without_ restricting the publicly accessible time range in general.

For example, a user might want to use the _stats_ endpoint in some custom script or for displaying [badges](https://shields.io) with a maximum time range of `last_year`, but at the same time have the card display only the past week. I'd find it cool to decouple this and allow to configure the card content independently from the underlying (generic) API endpoint.

PR #2734, opened 2023-05-18
Update vercel.json , to resolve an error

iDereje:


rickstaa:
@iDereje the max duration was set to 10 because most people don't have a pro-Vercel account. It, however, is needed for pro accounts to prevent extensive memory usage (see https://github.com/anuraghazra/github-readme-stats/issues/1416#issuecomment-950275476). We use https://github.com/anuraghazra/github-readme-stats/blob/master/.github/workflows/deploy-prep.py to set it for our pro vercel instance.

iDereje:
Oh , thanks I didn't knew about that.

PR #2742, opened 2023-05-22
Apply automatic changes

arwenjs:
asdsad

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -2,7 +2,7 @@
   "functions": {
     "api/*.js": {
       "memory": 128,
-      "maxDuration": 10
+      "maxDuration": 30
     }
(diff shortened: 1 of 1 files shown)


rickstaa:
@iDereje the max duration was set to 10 because most people don't have a pro-Vercel account. It, however, is needed for pro accounts to prevent extensive memory usage (see https://github.com/anuraghazra/github-readme-stats/issues/1416#issuecomment-950275476). We use https://github.com/anuraghazra/github-readme-stats/blob/master/.github/workflows/deploy-prep.py to set it for our Public (pro) Vercel instance.

iDereje:
Thanks for the Useful! information

PR #2745, opened 2023-05-24
build: add vercel to dev dependencies

rickstaa:
https://www.npmjs.com/package/vercel should be present in the `devDependencies` since it is needed for local development.


qwerty541 (commented):
Hey, @rickstaa! I have some uncertainty about the need to add this dependency to the project. It is that I personally use vercel as a global dependency due to the fact that it may be required for several different projects. That is, I don’t see much point in keeping it right inside the project. It is also worth considering that this is a very heavy dependency that will slow down continuous integration. Perhaps I'm missing something?

Issue #2737, opened 2023-05-19
Using layout parameter twice in top-langs gives toLowerCase not a function error


feikname:
EDIT: I was giving two layout parameters without noticing, ignore the PR reference! Nonetheless here's the bug report

If layout parameter is given two times such as in `https://github-readme-stats.vercel.app/api/top-langs/?username=feikname&layout=compact&layout=compact`

Then this happens

![image](https://github.com/anuraghazra/github-readme-stats/assets/4926869/6891a637-b5f7-40b3-8ebb-e406b7ff3f05)

qwerty541:
Remove extra layout parameter then. I'm not sure that it is a bug. What do you think @rickstaa ?

If we are going to fix it the source of problem is the following code string:

https://github.com/anuraghazra/github-readme-stats/blob/f9427b2a54bbab31aa7874da25a2e3227dbb4e77/src/cards/top-languages-card.js#L719

rickstaa:
@feikname thanks for your bug report. I fixed the issues your having in #2746.

rickstaa:
Fixed in https://github.com/anuraghazra/github-readme-stats/pull/2770.

PR #2746, opened 2023-05-24
fix: fix layout 'toLowerCase' bug

rickstaa:
This commit makes sure that the `layout?.toLowerCase is not a function` error is not thrown when users accidentally supply two layouts (see #2737)

### DIFF
--- a/src/cards/top-languages-card.js
+++ b/src/cards/top-languages-card.js
@@ -700,13 +700,16 @@ const renderTopLanguages = (topLangs, options = {}) => {
   let height = calculateNormalLayoutHeight(langs.length);
 
   let finalLayout = "";
-  if (layout === "pie") {
+  const layoutParsed = Array.isArray(layout)
+    ? layout[0]?.toLocaleLowerCase()
+    : layout?.toLocaleLowerCase();
+  if (layoutParsed === "pie") {
     height = calculatePieLayoutHeight(langs.length);
     finalLayout = renderPieLayout(langs, totalLanguageSize);
-  } else if (layout === "donut-vertical") {
+  } else if (layoutParsed === "donut-vertical") {
     height = calculateDonutVerticalLayoutHeight(langs.length);
     finalLayout = renderDonutVerticalLayout(langs, totalLanguageSize);
-  } else if (layout === "compact" || hide_progress == true) {
+  } else if (layoutParsed === "compact" || hide_progress == true) {
     height =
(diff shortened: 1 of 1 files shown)


qwerty541 (commented):
A few my thoughts on this change.

I think that this error (#2737) is user-side because according to the documentation we expect to receive a string as layout parameter, not an array. So it is users responsibility to send correct parameter value on server. But i agree that user should receive error not about `.toLowerCase()` function, error should be about parameter type mismatch. We should handle all parameter type mismatch errors and show them in user-friendly appearance.

And i think that we don't need to call `.toLowerCase()` function at all. Documentation contains all available layouts and it is users responsibility to reprint it without typos. We don't need to modify this value.

rickstaa:
> A few my thoughts on this change.
> 
> I think that this error (#2737) is user-side because according to the documentation we expect to receive a string as layout parameter, not an array. So it is users responsibility to send correct parameter value on server. But i agree that user should receive error not about `.toLowerCase()` function, error should be about parameter type mismatch. We should handle all parameter type mismatch errors and show them in user-friendly appearance.
> 
> And i think that we don't need to call `.toLowerCase()` function at all. Documentation contains all available layouts and it is users responsibility to reprint it without typos. We don't need to modify this value.

@qwerty541 I agree with you, but I assumed that handling this as an error would add more code compared to just dealing with the user error internally 😅. Feel free however to create a pull request for me to review.

qwerty541:
> > A few my thoughts on this change.
> > I think that this error (#2737) is user-side because according to the documentation we expect to receive a string as layout parameter, not an array. So it is users responsibility to send correct parameter value on server. But i agree that user should receive error not about `.toLowerCase()` function, error should be about parameter type mismatch. We should handle all parameter type mismatch errors and show them in user-friendly appearance.
> > And i think that we don't need to call `.toLowerCase()` function at all. Documentation contains all available layouts and it is users responsibility to reprint it without typos. We don't need to modify this value.
> 
> @qwerty541 I agree with you, but I assumed that handling this as an error would add more code compared to just dealing with the user error internally sweat_smile. Feel free however to create a pull request for me to review.

Hey, @rickstaa! I have opened pull request #2770 with alternate solution. In this pull request i detect incorrect layout parameter value input and show user-friendly error instead of confusing `layout?.toLowerCase is not a function`. And it is only 3 code string more than this pull request. Can you take a look?

rickstaa:
@qwerty541 Love your solution. Thanks. Merged into the main branch 🚀.

Issue #2735, opened 2023-05-19
NAN for language Percentage


CEbbinghaus:
### Describe the bug

When using the following url:
```
https://github-readme-stats.vercel.app/api/top-langs?theme=dark&size_weight=0&count_weight=1langs_count=10&layout=compact&username=cebbinghaus&hide=javascript,c,html,css&exclude_repo=srb2kart&aieBootstrap
```
The languages all show NaN for the Percentage. 

### Expected behavior

The correct Percentages should be shown

### Screenshots / Live demo link

Screenshot:
![image](https://github.com/anuraghazra/github-readme-stats/assets/31423391/9f0b0ceb-1ebe-45a0-828e-c83e5eb67c51)
Live demo:
![demo](https://github-readme-stats.vercel.app/api/top-langs?theme=dark&size_weight=0&count_weight=1langs_count=10&layout=compact&username=cebbinghaus&hide=javascript,c,html,css&exclude_repo=srb2kart&aieBootstrap)

CEbbinghaus:
The URL is malformed since there is a & missing after count_weight but I would expect it to spit out an error rather than NaN

rickstaa:
> The URL is malformed since there is a & missing after count_weight but I would expect it to spit out an error rather than NaN

@CEbbinghaus thanks for bringing this bug to our attention. It would be wonderful if you can create a pull request to resolve this issue. You can find the contribution guide at https://github.com/anuraghazra/github-readme-stats/blob/master/CONTRIBUTING.md.

manhhungpc:
Here is my fix for this issue, please see in [#2761](https://github.com/anuraghazra/github-readme-stats/pull/2761)

PR #2761, opened 2023-05-31
fix: validate top-langs query

manhhungpc:
This commit validate the query to make sure not display NaN percentage and throw an error when have an invalid query. For example: `/api/top-langs?&username=manhhungpc&theme=dark&size_weight=0&count_weight=1langs_count=10&layout=compact` where `&count_weight=1langs_count=10` is invalid query. See (#2735) for more details

### DIFF
--- a/api/top-langs.js
+++ b/api/top-langs.js
@@ -36,6 +36,37 @@ export default async (req, res) => {
   } = req.query;
   res.setHeader("Content-Type", "image/svg+xml");
 
+  const queriesName = [
+    "username",
+    "hide",
+    "hide_title",
+    "hide_border",
+    "card_width",
+    "title_color",
+    "text_color",
+    "bg_color",
+    "theme",
+    "cache_seconds",
+    "layout",
+    "langs_count",
+    "exclude_repo",
+    "size_weight",
+    "count_weight",
+    "custom_title",
+    "locale",
+    "border_radius",
+    "border_color",
+    "disable_animations",
+    "hide_progress",
+  ];
+
+  for (let [_, value] of Object.entries(req.query)) {
+    queriesName.map((name) => {
+      if (value.toString().includes(name))
+        return res.send(renderError("Something went wrong", "Invalid query"));
+    });
+  }
+
   if (blacklist.includes(username)) {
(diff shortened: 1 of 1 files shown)


api/top-langs.js
```diff
@@ -36,6 +36,37 @@ export default async (req, res) => {
...
+    "disable_animations",
+    "hide_progress",
+  ];
+
+  for (let [_, value] of Object.entries(req.query)) {
+    queriesName.map((name) => {
+      if (value.toString().includes(name))
+        return res.send(renderError("Something went wrong", "Invalid query"));
+    });
+  }
```
anuraghazra, line 68: Thanks for the PR @manhhungpc.

It will be better to stay a bit loose than this strict req.query checking. Right now even if people add an extraneous query or do a simple mistake this will throw error for the whole card. 
Not to mention it might also have a negative side effect - Imagine later down the line we remove or deprecate a query param, now people might not update the URL even if the query param is deprecated but this code will throw error and users will have no idea why it's failing. 

Instead it will be better if we don't throw instead if a query param is malformed we just ditch that param's value or default it to something sensible. 

Eg for the count_weight if its NaN we just disregard it.

rickstaa: > Thanks for the PR @manhhungpc.
> 
> It will be better to stay a bit loose than this strict req.query checking. Right now even if people add an extraneous query or do a simple mistake this will throw error for the whole card. Not to mention it might also have a negative side effect - Imagine later down the line we remove or deprecate a query param, now people might not update the URL even if the query param is deprecated but this code will throw error and users will have no idea why it's failing.
> 
> Instead it will be better if we don't throw instead if a query param is malformed we just ditch that param's value or default it to something sensible.
> 
> Eg for the count_weight if its NaN we just disregard it.

@manhhungpc I agree with @anuraghazra. Let's close this pr for now. Free free to create a new pull request that fixes the `NaN` bug explained in #2735 by ditching the incorrect parameter value and reverting back to default.

rickstaa: @manhhungpc on second. Maybe let's keep this open and apply the changes to this pull request?

PR #2767, opened 2023-06-04
Remove redundant async notation from several top languages card tests

qwerty541:


rickstaa:
You're correctly, it is using the render function directly 👍🏻.

rickstaa (approved):

Issue #2769, opened 2023-06-04
GitHub Stats are not Getting Updated


mukulwadhwa:
### Describe the bug

Right now, it is showing my GitHub stats as follows:

Total Stars Earned:
3
Total Commits (2023):
41
Total PRs:
0
Total Issues:
2
Contributed to (last year):
0

### Expected behavior

It should have updated following stats: 

Total Commits (2023):
41
Total PRs:
0

qwerty541:
By default stats card shows only public contributions. I do not want to repeat myself, see the answer here https://github.com/anuraghazra/github-readme-stats/issues/2768#issuecomment-1576106113

rickstaa:
> By default stats card shows only public contributions. I do not want to repeat myself, see the answer here [#2768 (comment)](https://github.com/anuraghazra/github-readme-stats/issues/2768#issuecomment-1576106113)

I feel you. This question is being asked a lot. I am starting to think we need to update the documentation. I think https://github.com/anuraghazra/github-readme-stats/pull/2736 will already remove some confusion 🤔.

Issue #2768, opened 2023-06-04
The card doesn't show all my commits and pr's


PigeonTheDev:
### Describe the bug

The card describes that Ihave 0 pr's even though I opened 10 prs just  last week, And I' sure I don't have just 30 commits

qwerty541:
The card shows correct data because by default it show only your public contributions. I consider that 10 PRs you are talking about are private contributions. To show your private contributions on stats card you should [deploy your own vercel instance](https://github.com/anuraghazra/github-readme-stats#deploy-on-your-own) using your own GitHub API token which allowed to see your private data.

PR #2776, opened 2023-06-06
Add note about private contributions into readme

qwerty541:
Related to #2768 and #2769

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -108,6 +108,9 @@ Change the `?username=` value to your GitHub username.
 ```
 
 > **Note**
+> By default stats card shows contributions only from public repositories. To show your private contributions you should [deploy you own instance](#deploy-on-your-own) using your own GitHub API token which allowed to see your private data.
+
+> **Note**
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -107,6 +107,9 @@ Change the `?username=` value to your GitHub username.
 [![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=anuraghazra)](https://github.com/anuraghazra/github-readme-stats)
 ```
 
+> **Note**
```
rickstaa, line 110: Great addition. I would, however, make it a **Warning** admonition since it is important to know. Further, maybe we can move it under the other note. What do you think?

rickstaa: Or do you think a warning is too aggressive? Then your version is okay 👍🏻.

qwerty541: Yeah, i agree that **Warning** will be better there, i have updated the code. Regarding the second question, I think information about private contributions is more important than available ranks details, at least because of the large number of issues that users create on this.

rickstaa: > Yeah, i agree that **Warning** will be better there, i have updated the code. Regarding the second question, I think information about private contributions is more important than available ranks details, at least because of the large number of issues that users create on this.

I agree with you on that! Thanks for the update. BTW, welcome to the team 🚀!

rickstaa (approved):

readme.md
qwerty541: > BTW, welcome to the team 🚀!

Thanks! Feel free to send me pull requests for review which you want to merge. Now you don't have to wait a while to get an approvement, probably 😄

rickstaa: Nice 🚀!

PR #2792, opened 2023-06-08
Update readme.md about maxDuration

ForsakenRei:
I think we can remove this warning section since that value in  `.json` has been changed to 10 by default nowadays?

Also removed related section in self deployment guide.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -578,6 +578,9 @@ Since the GitHub API only allows 5k requests per hour, my `https://github-readme
 > **Note**
 > Since [#58](https://github.com/anuraghazra/github-readme-stats/pull/58), we should be able to handle more than 5k requests and have fewer issues with downtime :grin:.
 
+> **Note**
+> If you are on the [Pro (i.e. paid)](https://vercel.com/pricing) Vercel plan, the [maxDuration](https://vercel.com/docs/concepts/projects/project-configuration#value-definition) value found in the [Vercel.json](https://github.com/anuraghazra/github-readme-stats/blob/master/vercel.json) can be increased when your Vercel instance frequently times out during the card request. You are advised to keep this value lower than `30` seconds to prevent high memory usage.
+
 [![Deploy to Vercel](https://vercel.com/button)](https://vercel.com/import/project?template=https://github.com/anuraghazra/github-readme-stats)
 
 <details>
--- a/readme.md
+++ b/readme.md
@@ -590,16 +590,15 @@ Since the GitHub API only allows 5k requests per hour, my `https://github-readme
     ![](https://files.catbox.moe/b9oxey.png)
 4.  Sign in to GitHub and allow access to all repositories if prompted.
 5.  Fork this repo.
-6.  After forking the repo, open the [`vercel.json`](https://github.com/anuraghazra/github-readme-stats/blob/master/vercel.json#L5) file and change the `maxDuration` field to `10`.
-7.  Go back to your [Vercel dashboard](https://vercel.com/dashboard).
-8.  To import a project, click the `Add New...` button and select the `Project` option.
+6.  Go back to your [Vercel dashboard](https://vercel.com/dashboard).
+7.  To import a project, click the `Add New...` button and select the `Project` option.
     ![](https://files.catbox.moe/3n76fh.png)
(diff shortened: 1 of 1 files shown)


qwerty541 (commented):
@ForsakenRei Looks like you are right and we no longer need to keep this warning section, but i think that probably it can be useful to add inverse warning for pro plan users to remind them about ability to increase maximum execution time. What do you think @rickstaa?

ForsakenRei:
That's a good point. Though I'm not sure if there is any benefit of increasing the max execution time, like more frequently data refresh?

qwerty541:
> That's a good point. Though I'm not sure if there is any benefit of increasing the max execution time, like more frequently data refresh?

Parameter `maxDuration` is an integer defining how long your serverless function should be allowed to run on every request in seconds. I think that increasing max execution time make sense when your vercel instance experiencing high load and you are using several different GitHub API tokens. Currently the application works the following way: it iterates through provided API tokens array and try to fetch card data until one of requests succeed. When most of your API tokens hit the limiter it start taking a while.

ForsakenRei:
> > That's a good point. Though I'm not sure if there is any benefit of increasing the max execution time, like more frequently data refresh?
> 
> Parameter `maxDuration` is an integer defining how long your serverless function should be allowed to run on every request in seconds. I think that increasing max execution time make sense when your vercel instance experiencing high load and you are using several different GitHub API tokens. Currently the application works the following way: it iterates through provided API tokens array and try to fetch card data until one of requests succeed. When most of your API tokens hit the limiter it start taking a while.

I see, I'm open to this suggestion for sure.

rickstaa:
@ForsakenRei and @qwerty541 the reason we used to have a `maxDuration` of 30 on the public instance was the following:

> Open the vercel.json file and change the maxDuration to 10.
>
> It's not a bug, it was intentionally reduced to 30s because by default on PRO plan it sets to 1min which is too high and was causing too much memory usage. but for non-pro users the duration can only be <10s thus 30s fails.
>
> _Originally posted by @anuraghazra in https://github.com/anuraghazra/github-readme-stats/issues/1416#issuecomment-950275476_

This behaviour was simplified by @Zo-Bro-23 who created a GitHub action that changes the `maxDuration` on our public instance during deployment. This was done since more people are on the free Vercel plan than the Pro Vercel plan. 

So, you're right @ForsakenRei, the message in the readme should be changed. I, however, would change it to something along the line of:

> **Note:**
> If you are on the [Pro (i.e. paid)](https://vercel.com/pricing) Vercel plan, the `maxDuration` value found in the [Vercel.json](https://github.com/anuraghazra/github-readme-stats/blob/master/vercel.json) file might cause high memory usage. In this case, you are advised to change it to `30` seconds.

rickstaa:
> @ForsakenRei and @qwerty541 the reason we used to have a `maxDuration` of 30 on the public instance was the following:
> 
> > Open the vercel.json file and change the maxDuration to 10.
> > It's not a bug, it was intentionally reduced to 30s because by default on PRO plan it sets to 1min which is too high and was causing too much memory usage. but for non-pro users the duration can only be <10s thus 30s fails.
> > _Originally posted by @anuraghazra in [#1416 (comment)](https://github.com/anuraghazra/github-readme-stats/issues/1416#issuecomment-950275476)_
> 
> This behaviour was simplified by @Zo-Bro-23 who created a GitHub action that changes the `maxDuration` on our public instance during deployment. This was done since more people are on the free Vercel plan than the Pro Vercel plan.
> 
> So, you're right @ForsakenRei, the message in the readme should be changed or removed.

I updated this message to remove the suggested warning, as it was incorrect. What I meant to suggest was:

 I, however, would change it to something along the line of:

> **Note**
> If you are on the [Pro (i.e. paid)](https://vercel.com/pricing) Vercel plan, the [maxDuration](https://vercel.com/docs/concepts/projects/project-configuration#value-definition) value found in the [Vercel.json](https://github.com/anuraghazra/github-readme-stats/blob/master/vercel.json) can be increased when your Vercel instance frequently times out during the card request. You are advised to keep this value lower than `30` seconds to prevent high memory usage.

@qwerty541 I added this suggestion to this pull request. Please review 🙏🏻.

rickstaa (approved):

qwerty541 (approved):

PR #2799, opened 2023-06-10
Remove restrictedContributionsCount from GraphQL query and tests

qwerty541:


qwerty541 (commented):
Since `count_private` option was removed this code is no longer used.

rickstaa (approved):

rickstaa:
Great eye for detail 👍🏻.

Issue #2793, opened 2023-06-08
Personal Instance with Repo Access PAT Cannot Read Yearly Commit Correctly


ForsakenRei:
### Describe the bug

Update 6/10: Current situation explained in https://github.com/anuraghazra/github-readme-stats/issues/2793#issuecomment-1585832224

----

Sorry I know this have been asking multiple times recently, I have been using my own vercel instance for months and everything worked well until I synced my fork yesterday and remove `count_private` was inplemented. Yesterday it shows 300+ commits but today it's 200+. 

Counted from my contribution activity it should be 300ish in 2023, if it is last year from June 2022 to June 2023 it will be 400ish.

Top language seems working properly since the percentage didn't change much from yesterday but status card only shows public commits now? If I add `include_all_commits=true` the commit number(1.3K for all years) seems contain all my commits regardless of private or public repo.

I recreated a vercel instance based on [38c8a7a](https://github.com/ForsakenRei/github-readme-stats/commit/38c8a7a1e4fea71409f7c4834f7c7b1b8cd695ca) with a new PAT granted repo access, but the behavior is still the same.

Am I missing any argument?

### Expected behavior

Status card includes commits to private repo as well.

### Screenshots / Live demo link
![include_all_commits](https://github-readme-stats-murex-five-78.vercel.app/api?username=forsakenrei&show_icons=true&theme=tokyonight&include_all_commits=true)

![normal](https://github-readme-stats-murex-five-78.vercel.app/api?username=forsakenrei&show_icons=true&theme=tokyonight)

https://github.com/ForsakenRei/ForsakenRei

From official instance
![include_all_commits](https://github-readme-stats.vercel.app/api?username=forsakenrei&show_icons=true&theme=tokyonight&include_all_commits=true)

![normal](https://github-readme-stats.vercel.app/api?username=forsakenrei&show_icons=true&theme=tokyonight)

salesp07:
Same problem here. They should revert the commit where `count_private` was removed until this is sorted out.

rickstaa:
@ForsakenRei I just checked your stats, and for the `include_all_commits=false` case it looks like you did something wrong while deploying your [own Vercel instance](https://github.com/anuraghazra/github-readme-stats#deploy-on-your-own). Please ensure that the PAT has the rights to read data from your private repositories. You can check this out yourself using the https://docs.github.com/en/graphql/overview/explorer:

```graphql
query userInfo($login: String!) {
  user(login: $login) {
[...]

PR #2805, opened 2023-06-10
docs: add  PAT permission to docs

rickstaa:
This pull request changes the README so that users are aware that they also need to set the `user` permission when creating the PAT when they want their private commits to show up.


ForsakenRei:
Ref -> https://github.com/anuraghazra/github-readme-stats/issues/2793

qwerty541 (approved):

PR #2829, opened 2023-06-15
Revert "Use default json import"

rickstaa:
Reverts anuraghazra/github-readme-stats#2820


rickstaa:
@qwerty541 I could not find why the JSON import syntax is still failing under node 18. I think the easiest way to fix is to revert this change.

rickstaa:
Looks like type assertions are still experimental in node v18.x even though they were planned (see https://nodejs.org/dist/latest-v18.x/docs/api/esm.html#import-assertions). We therefore still need to use the workarround.

rickstaa:
More information on why this happens can be found at https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/.

PR #1404, opened 2021-10-21
Added pr reviews to stats

Marcholio:
Fixes: #1394 

Added code reviews to stats card:
Screenshot 2021-10-21 at 17 38 40

What do you think about this feature? They do not affect the rank at the moment, that's something that should be maybe added later. Also translations just default to english.

PR #2834, opened 2023-06-16
Stats card: Reorder total reviews count position

qwerty541:
Hey, @rickstaa! I see that you reverted this change while merging #1404. I deliberately placed the number of reviews next to the number of pull requests, because in my opinion the card looks much prettier in this way and, in general, the arrangement of these elements side by side is more logical due to their close relationship with each other. Do not you agree?


rickstaa:
Ah I forgot the order there effected the card order 😅.

rickstaa (approved):

PR #2849, opened 2023-06-17
Stats card: show started and answered discussions stats

qwerty541:
изображение


rickstaa (changes requested):
@qwerty541 I think you made a perfect pull request. The only thing I would do is to split discussions into discussions and answered_discussion. I, for example, don't want to show the total discussions on the card but would like to show the number of answered discussions now that GitHub removed that from their UI.

qwerty541 (commented):
> @qwerty541 I think you made a perfect pull request. The only thing I would do is to split discussions into discussions and answered_discussion. I, for example, don't want to show the total discussions on the card but would like to show the number of answered discussions now that GitHub removed that from their UI.

@rickstaa You are right that the started and answered discussions should be splitted. I have update the code, please check it.

rickstaa (approved):

PR #2852, opened 2023-06-18
refactor: remove vscode 'formatOnSave'

rickstaa:
This commit ensures that the `formatOnSave` variable is not set. This
needs to be done since people have different prefered formatters. With `formatOnSave`
included this will cause code to be formatted on save with different
formatters than prettier.


qwerty541 (commented):
See: https://github.com/anuraghazra/github-readme-stats/pull/2844#issuecomment-1596445895

Issue #2846, opened 2023-06-17
Take into account user reviewed PRs count during rank calculation and update other weights


qwerty541:
Since stats card able to show users total reviewed PRs count after merging #1404 i suggest to consider possibility of taking into account this value during rank calculation. I think that any experienced programmer at a certain stage begins not only to write code, but also to review someone else's code, it would be logical that in order to achieve the highest rank, it would be necessary to score a certain number of reviews. What do you think about this @francois-rozet @rickstaa?

francois-rozet:
I am also favorable to take reviews into account. However, the weight should be fairly low (same as Issues maybe). I also have no idea of a good "target" amount of reviews.

rickstaa:
> I am also favorable to take reviews into account. However, the weight should be fairly low (same as Issues maybe). I also have no idea of a good "target" amount of reviews.

@qwerty541 if the stats master, aka @francois-rozet 👑, thinks it's a good idea, I'm also okay with it 👍🏻. 

@qwerty541, maybe you can give us an example PR with a review weight and mean. For the weight, I think this should be `1` same as issues and followers. For the mean value, at first, though I think it should be higher than PRs since reviews are easier to do. However, I noticed that @francois-rozet set issues lower than PRs. I think this is because the issue weight is lower, and the weight and mean are connected 🤔 . @francois-rozet, do you have a suggestion for the review mean?

francois-rozet:
Actually, the median number of issues was kind of arbitrary. Some decs never submit issues but submit PRs that fix issues. You often see maintainers with more PRs than issues.

rickstaa:
> Actually, the median number of issues was kind of arbitrary. Some devs never submit issues but submit PRs that fix issues. You often see maintainers with more PRs than issues. That's why I set it to a lower median, but it is not linked to the weight.
> 
> For the reviews, I think it should not be based on the easiness of the task, but on the actual frequency of reviews. Would a typical dev have more PRs or reviews? I don't know honestly

@francois-rozet Ah, thanks for the explanation. I just created some scripts that can help us retrieve the mean reviews done by the 52k stargazers of our repository. You can check it out [here](https://github.com/rickstaa/get-stargazers-info). I think if we let this run for a while we would have the mean number of reviews of our user base. Would this help you define a `reviews` weight and mean value?

qwerty541:
[...]

PR #2857, opened 2023-06-19
Ranks: Take into account user reviewed PRs count

qwerty541:
Related to #2846

TODOS:

- [ ] Discuss reviews median and weight values
- [x] Fix e2e test
- [x] ? update docs


francois-rozet (approved):
LGTM

src/calculateRank.js
```diff
@@ -36,6 +38,8 @@ function calculateRank({
     PRS_WEIGHT = 3;
   const ISSUES_MEDIAN = 25,
     ISSUES_WEIGHT = 1;
+  const REVIEWS_MEDIAN = 10,
```
rickstaa, line 41: @francois-rozet do you think these are okay given that less than 10% of users have more than 4 reviews?

francois-rozet: Oh good catch, yes the median is too high. Maybe 1 or 2?

rickstaa, line 41:
```suggestion
  const REVIEWS_MEDIAN = 2,
```

rickstaa (changes requested):

src/calculateRank.js
rickstaa: @francois-rozet, thanks for the quick response! I would love to have it at `20` since I want people to do more PR reviews :smiling_imp:. But given the data we aquired, that would be unfair :rofl:. Let's do `2` 👍🏻.

qwerty541: I have updated reviews median value and tests. Feel free to merge.

rickstaa (approved):

rickstaa:
@qwerty541 thanks again for the great PR!

@francois-rozet feel free to open a pull request with the updated weights when you have the time. No hurries :blue_heart:.

PR #2882, opened 2023-06-24
refactor: object shorthand

mxdvl:
For your consideration, a [shorthand property in object initialisation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions).

PR #2885, opened 2023-06-27
Eslint integration

qwerty541:
Related to https://github.com/anuraghazra/github-readme-stats/pull/2882#issuecomment-1606993478

### DIFF
--- a/src/getStyles.js
+++ b/src/getStyles.js
@@ -1,4 +1,4 @@
-// @ts-check
+// @ts-nocheck
 /**
  * Calculates progress along the boundary of the circle i.e it's circumference.
  *
@@ -75,6 +75,7 @@ const getAnimations = () => {
  * @returns {string} Card CSS styles.
  */
 const getStyles = ({
+  // eslint-disable-next-line no-unused-vars
   titleColor,
   textColor,
   iconColor,
--- a/src/getStyles.js
+++ b/src/getStyles.js
@@ -1,4 +1,5 @@
-// @ts-nocheck
+// @ts-check
+
 /**
  * Calculates progress along the boundary of the circle i.e it's circumference.
  *
--- a/tests/renderWakatimeCard.test.js
+++ b/tests/renderWakatimeCard.test.js
@@ -7,7 +7,7 @@ import { expect, it, describe } from "@jest/globals";
 
 describe("Test Render Wakatime Card", () => {
   it("should render correctly", () => {
-    // renderWakatimeCard(wakaTimeData.data);
+    // const card = renderWakatimeCard(wakaTimeData.data);
     expect(getCardColors).toMatchSnapshot();
   });
 
--- a/tests/renderWakatimeCard.test.js
+++ b/tests/renderWakatimeCard.test.js
@@ -3,10 +3,11 @@ import "@testing-library/jest-dom";
 import { renderWakatimeCard } from "../src/cards/wakatime-card.js";
 import { getCardColors } from "../src/common/utils.js";
 import { wakaTimeData } from "./fetchWakatime.test.js";
+import { expect, it, describe } from "@jest/globals";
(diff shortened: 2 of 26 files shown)


qwerty541 (commented):
@rickstaa This is basic integration of eslint. As you see most of rules inside `eslintrc.json` under comments, enabled only `object-shorthand` and default. We can gradually uncomment rules and improve code quality. I also added checking lints on continuous integration and pre commit hook.

tests/renderWakatimeCard.test.js
```diff
@@ -3,10 +3,11 @@ import "@testing-library/jest-dom";
 import { renderWakatimeCard } from "../src/cards/wakatime-card.js";
 import { getCardColors } from "../src/common/utils.js";
 import { wakaTimeData } from "./fetchWakatime.test.js";
+import { expect, it, describe } from "@jest/globals";
 
 describe("Test Render Wakatime Card", () => {
   it("should render correctly", () => {
-    const card = renderWakatimeCard(wakaTimeData.data);
```
rickstaa, line 9: @qwerty541, why was this line disabled?

src/getStyles.js
```diff
@@ -1,4 +1,4 @@
-// @ts-check
+// @ts-nocheck
```
rickstaa, line 1: @qwerty541, did you temporarily disable ts here because there are too many errors to fix?

rickstaa (changes requested):
See my comments. Appart from these two, I think this PR is good to go.

src/getStyles.js
qwerty541: There is no errors inside this file. And i am sure that i did not changed this thing manually. Idk what happened, i revert this change.

tests/renderWakatimeCard.test.js
qwerty541: Variable `card` unused in this test, this line do nothing.

qwerty541: We can remove it, or leave, maybe someone will remember why it was added. As you see it does not affect on tests running result.

src/getStyles.js
qwerty541: Nice catch :+1:

anuraghazra (approved):

tests/renderWakatimeCard.test.js
rickstaa: This test needed to be completed. We already have a test of the `getCardColors` function:

https://github.com/anuraghazra/github-readme-stats/blob/2c26329e136dade2b1b42f5ba6b495632731363c/tests/utils.test.js#L54-L96

With line 10 not commented out, the test will fail if the waketime rendering throws a syntax error. It would be better to remove line 11 and write a proper test as we have in the language and stats cards:

https://github.com/anuraghazra/github-readme-stats/blob/7c104cf8c58d55cd76a07806760a3455dfd40364/tests/renderTopLanguages.test.js#L31-L59

https://github.com/anuraghazra/github-readme-stats/blob/45941f6c70a56901652b9d8941ad1b365f83fa87/tests/renderStatsCard.test.js#L16-L32

Alternatively, we can also use a snapshot comparison for the output of the `renderWakatimeCard` function. We can also choose to use a snapshot test for the Wakatime card.

Therefore, I suggest reverting the change to 10 and adding an issue or pull request to improve this test.

rickstaa (changes requested):
See my suggestion

tests/renderWakatimeCard.test.js
qwerty541: @rickstaa I have opened an issue #2906 to not forget about this test improvement. I think it's better for now to keep this line commented because eslint shows warning about no-used variable.

rickstaa (approved):

PR #2903, opened 2023-07-02
ci: fix id-token 'read' bug

rickstaa:
This PR fixes a bug that is present when the `read` option is used for the `id-token` permission (see https://github.com/github/docs/issues/26481).


rickstaa:
Closed in favor of 2904.

PR #2926, opened 2023-07-09
revise chinese version readme

seanxuu:
There is a revised version of the Chinese readme.md. Because a picture in the doc is not valid. And I add a note based on https://github.com/anuraghazra/github-readme-stats/discussions/2911#discussioncomment-6354828


rickstaa:

PR #2945, opened 2023-07-15
CI: Avoid test workflow duplicate

qwerty541:
Trying to avoid test workflow duplicate

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,11 @@
 name: Test
-on: [push, pull_request]
+on:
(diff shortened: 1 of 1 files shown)


rickstaa (approved):

rickstaa:
@qwerty541 Your right. Based on the documentation the `pull_request` event is triggered when a PR is `opened`, `synchronize`, or `reopened` (see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). This should be enough.

PR #2949, opened 2023-07-18
Refactor: Fixed docstring for flexLayout function

qwerty541:


src/common/utils.js
```diff
@@ -173,8 +173,8 @@ const request = (data, headers) => {
  * @param {object} props Function properties.
  * @param {string[]} props.items Array of items to layout.
  * @param {number} props.gap Gap between items.
- * @param {number[]?=} props.sizes Array of sizes for each item.
- * @param {"column" | "row"?=} props.direction Direction to layout items.
+ * @param {number[]=} props.sizes Array of sizes for each item.
```
rickstaa, line 176:
```suggestion
 * @param {"column" | "row"=} props.direction Direction to layout items.
```

src/common/utils.js
```diff
@@ -173,8 +173,8 @@ const request = (data, headers) => {
...
+ * @param {"column" | "row"=} props.direction Direction to layout items.
```
rickstaa, line 177:
```suggestion
 * @param {number[]=} props.sizes Array of sizes for each item.
```

rickstaa: Maybe we can swap the order so that it is the same as in the function description?

rickstaa (changes requested):

src/common/utils.js
qwerty541: Good suggestion, i have updated the code, check it please.

rickstaa (approved):

PR #2957, opened 2023-07-21
docs: fix broken admonitions

rickstaa:
This pull request ensures that the admonitions are in line with the new GFM specifications (see https://github.com/orgs/community/discussions/16925#discussioncomment-6506860).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,8 @@
 {
     "markdown.extension.toc.levels": "1..3",
     "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode" 
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "[markdown]": {
+        "editor.defaultFormatter": "esbenp.prettier-vscode"
+    }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,4 @@
     "markdown.extension.toc.levels": "1..3",
     "editor.formatOnSave": true,
     "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "[markdown]": {
-        "editor.defaultFormatter": "esbenp.prettier-vscode"
-    }
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,20 +40,20 @@ _(make sure you already have a [Vercel](https://vercel.com/) account)_
 8.  Run the command `vercel dev` to start a development server at <https://localhost:3000>.
 9.  The cards will then be available from this local endpoint (i.e. `https://localhost:3000/api?username=anuraghazra`).
 
-> **Note**
+> [!NOTE]\
 > You can debug the package code in [Vscode](https://code.visualstudio.com/) by using the [Node.js: Attach to process](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_setting-up-an-attach-configuration) debug option. You can also debug any tests using the [VSCode Jest extension](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest). For more information, see https://github.com/jest-community/vscode-jest/issues/912.
 
 ## Themes Contribution
 
 GitHub Readme Stats supports custom theming, and you can also contribute new themes!
 
-> **Note**
+> [!NOTE]\
 > If you are contributing your theme just because you are using it personally, then you can [customize the looks](./readme.md#customization) of your card with URL params instead.
 
-> **Note**
+> [!NOTE]\
 > Keep in mind that we already have a vast collection of different themes. To keep their number manageable, we began to add only themes supported by the community. Your pull request with theme addition will be merged once we get enough positive feedback from the community in the form of thumbs up (see [#1935](https://github.com/anuraghazra/github-readme-stats/issues/1935#top-themes-prs)). Remember that you can also support themes of other contributors that you liked to speed up their merge.
 
-> **Note**
+> [!NOTE]\
 > Before submitting pull request, please make sure that your theme pass WCAG 2.0 level AA constrast ration test. You can use [this tool](https://webaim.org/resources/contrastchecker/) to check it.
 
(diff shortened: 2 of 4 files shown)


qwerty541 (commented):
@rickstaa Nice improvement, but i see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.

rickstaa:
> @rickstaa Nice improvement, but i see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.

These changes were added by Vscode when I saved the file using `Ctrl + S` so assumed they were the ones you put in the config file 🤔. Strange.

qwerty541:
> > @rickstaa Nice improvement, but i see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.
> 
> These changes were added by Vscode when I saved the file using `Ctrl + S` so assumed they were the ones you put in the config file 🤔. Strange.

We have only one config related to markdown: https://github.com/anuraghazra/github-readme-stats/blob/a3b8a7856cb8a861a3267286ea76c80894279007/.vscode/settings.json#L2

Most likely you have some configs on user's level, you should move them into workspace configs or discard changes.

rickstaa:
> > > @rickstaa Nice improvement, but I see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.
> > 
> > 
> > These changes were added by Vscode when I saved the file using `Ctrl + S` so assumed they were the ones you put in the config file thinking. Strange.
> 
> We have only one config related to markdown:
> 
> https://github.com/anuraghazra/github-readme-stats/blob/a3b8a7856cb8a861a3267286ea76c80894279007/.vscode/settings.json#L2
> 
> Most likely you have some configs on user's level, you should move them into workspace configs or discard changes.

We need to choose a default format for both JS and markdown 👍🏻. I have three on my system. Which one do you want to use for markdown? Also prettier?

qwerty541:
> > > > @rickstaa Nice improvement, but I see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.
> > > 
> > > 
> > > These changes were added by Vscode when I saved the file using `Ctrl + S` so assumed they were the ones you put in the config file thinking. Strange.
> > 
> > 
> > We have only one config related to markdown:
> > https://github.com/anuraghazra/github-readme-stats/blob/a3b8a7856cb8a861a3267286ea76c80894279007/.vscode/settings.json#L2
> > 
> > Most likely you have some configs on user's level, you should move them into workspace configs or discard changes.
> 
> We need to choose a default format for both JS and markdown 👍🏻. I have three on my system. Which one do you want to use for markdown? Also prettier?

Does `yzhang.markdown-all-in-one` extension provide formatter for markdown as well? We can keep prettier as formatter for JS and set another formatter for markdown if it is better at something.

rickstaa:
> > > > > @rickstaa Nice improvement, but I see there a lot of other changes like replacing `-` symbols with `*` in all kinds of lists and backslashes before all `&`, `[` and `]` symbols. I assume that it was not manual chacnges, if we're going to use it, workspace settings should be updated. Otherwise if i will do some changes in this file in future and press `Ctrl + S`, these changes most likely will be reverted.
> > > > 
> > > > 
> > > > These changes were added by Vscode when I saved the file using `Ctrl + S` so assumed they were the ones you put in the config file thinking. Strange.
> > > 
> > > 
> > > We have only one config related to markdown:
> > > https://github.com/anuraghazra/github-readme-stats/blob/a3b8a7856cb8a861a3267286ea76c80894279007/.vscode/settings.json#L2
> > > 
> > > Most likely you have some configs on user's level, you should move them into workspace configs or discard changes.
> > 
> > 
> > We need to choose a default format for both JS and markdown 👍🏻. I have three on my system. Which one do you want to use for markdown? Also prettier?
> 
> Does `yzhang.markdown-all-in-one` extension provide formatter for markdown as well? We can keep prettier as formatter for JS and set another formatter for markdown if it is better at something.

Yea, the incorrect formatting came from that extension. I now updated the `settings.json` to prevent that from happening. Let's stay with Prettier since it is the most used and most versatile it also is the one we already use.

.vscode/settings.json
```diff
@@ -1,5 +1,8 @@
 {
     "markdown.extension.toc.levels": "1..3",
     "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode" 
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "[markdown]": {
+        "editor.defaultFormatter": "esbenp.prettier-vscode"
+    }
```
qwerty541, line 7: Does `editor.defaultFormatter` option affect on the whole workspace including all file extensions? I'm not sure if we need to define same option for markdown files separately.

qwerty541 (approved):

.vscode/settings.json
qwerty541: @rickstaa I resolved this

Issue #2637, opened 2023-04-23
Change error card


PencilNavigator:
### Is your feature request related to a problem? Please describe.

![pic](https://user-images.githubusercontent.com/96688819/212421796-dd6535d9-7607-4343-ab61-7a2e8015272b.png)
as you see, in the error page card it tells users to make a issue in this github repository. This can be misleading for some users because they just go straight ahead fileing a issue instead of searching all pass issues.

i used the keyword "https://tiny.one/readme-stats" to search, and i found about 15 duplicate issues all because of PAT limit and the error card was shown, telling users to file a issue here.
![pic](https://user-images.githubusercontent.com/85282140/233851938-aae3aaea-1b13-4fc7-9904-ba53ca08cf7b.png)

### Describe the solution you'd like

Instead of telling users to file a issue straight, i think we should refer them to a documentation/wiki page that includes most of the common errors and their fixes. Then in that page, mention if non of the above fixes your issue, then go ahead and file a issue.

### Describe alternatives you've considered

if the above solution is too hard for the maintainers to do, then just change https://tiny.one/readme-stats redirect link to readme and change "file an issue" to "check readme.md"

### Additional context

N/A

rickstaa:
@PencilNavigator good idea. You are welcome to create a pull request to improve the behaviour 👍🏻. We now store the common error codes in https://github.com/anuraghazra/github-readme-stats/issues/1772 in the hope people will check that, but you are welcome to suggest a better place.

PencilNavigator:
> @PencilNavigator good idea. You are welcome to create a pull request to improve the behaviour 👍🏻. We now store the common error codes in #1772 in the hope people will check that, but you are welcome to suggest a better place.

why not store it in Github wiki? imo storing it in a issue isn't the best idea to do as issues are for reports, not documentation.

rickstaa:
> > @PencilNavigator good idea. You are welcome to create a pull request to improve the behaviour 👍🏻. We now store the common error codes in #1772 in the hope people will check that, but you are welcome to suggest a better place.
> 
> why not store it in Github wiki? imo storing it in a issue isn't the best idea to do as issues are for reports, not documentation.

That would also be a good place. The Wiki, however, has to be enabled by @anuraghazra if we want to do this. I don't have those permissions 😅.

anuraghazra:
> why not store it in Github wiki?

[...]

PR #2962, opened 2023-07-21
Switch to data-driven percentiles

francois-rozet:
This PR modifies the ranking system to use the actual quantiles (see #2857) of the statistics. This leads to significantly different, but generally higher, ranks for users. In particular, it is now impossible to be at C rank (minimum is C+). An alternative to lower the ranks would be to filter out inactive users before computing the quantiles.

What do you think @rickstaa, @qwerty541?

### DIFF
--- a/src/calculateRank.js
+++ b/src/calculateRank.js
@@ -1,11 +1,90 @@
-function exponential_cdf(x) {
-  return 1 - 2 ** -x;
+function score(x, quantiles) {
+  const i = quantiles.findIndex((q) => x < q);
+
+  if (i == 0) {
+    return 0.0;
+  } else if (i == -1) {
+    return 1.0;
+  }
+
+  const a = quantiles[i - 1];
+  const b = quantiles[i];
+
+  return ((x - a) / (b - a) + i - 1) / (quantiles.length - 1);
 }
 
-function log_normal_cdf(x) {
-  // approximation
-  return x / (1 + x);
-}
+const QUANTILES = {
+  commits: [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 3, 3,
+    4, 4, 5, 6, 7, 7, 8, 9, 10, 11, 12, 13, 15, 16, 17, 19, 20, 22, 23, 25, 27,
+    29, 31, 33, 35, 38, 40, 43, 45, 48, 51, 54, 57, 60, 64, 67, 71, 76, 80, 85,
+    89, 94, 99, 105, 111, 118, 125, 132, 140, 147, 155, 164, 173, 184, 195, 207,
+    220, 233, 249, 265, 284, 304, 326, 353, 380, 411, 451, 495, 545, 611, 691,
+    794, 933, 1195, 1704, 9722,
+  ],
+  all_commits: [
+    0, 0, 0, 0, 2, 4, 8, 11, 15, 19, 23, 27, 32, 36, 41, 45, 50, 55, 60, 65, 71,
+    76, 82, 87, 93, 99, 105, 111, 117, 124, 131, 137, 145, 151, 159, 166, 174,
+    182, 190, 198, 207, 215, 225, 234, 244, 253, 264, 274, 285, 296, 306, 318,
+    330, 342, 355, 368, 382, 396, 409, 424, 440, 457, 475, 493, 512, 531, 551,
+    570, 593, 618, 643, 667, 695, 723, 752, 784, 815, 857, 893, 934, 984, 1037,
+    1094, 1152, 1217, 1289, 1379, 1475, 1576, 1696, 1851, 2023, 2232, 2480,
+    2835, 3242, 3885, 4868, 6614, 11801, 792319,
+  ],
+  prs: [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1,
+    1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7,
+    8, 8, 9, 10, 10, 11, 11, 12, 13, 14, 14, 15, 16, 17, 18, 19, 20, 21, 23, 24,
+    25, 27, 28, 30, 32, 34, 36, 38, 40, 43, 46, 50, 53, 57, 61, 65, 70, 76, 83,
+    90, 99, 110, 123, 139, 159, 185, 219, 273, 360, 562, 2291,
+  ],
+  issues: [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3,
+    4, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7, 8, 8, 9, 10, 10, 11, 12, 12, 13, 14, 15,
+    16, 17, 19, 20, 21, 23, 24, 26, 28, 30, 32, 35, 38, 41, 45, 49, 54, 59, 66,
+    73, 82, 92, 106, 123, 150, 186, 255, 409, 1590,
+  ],
+  reviews: [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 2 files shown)


qwerty541 (commented):
In general, the fact that it is now impossible to get a C rank is quite logical. If you look closely at the quantiles, you can see that at the beginning there are zeros everywhere. I would even suggest that a C+ rank would be no less rare, since a new user with empty statistics gets a percentile of 78.26086956521738 in the new system, and the threshold for getting a B rank is only 80. I think that the thresholds need to be changed, at least for the latest ranks. My suggestion is the following:

```js
const THRESHOLDS = [1, 12.5, 25, 37.5, 45, 52.5, 60, 67.5, 75];
```

@francois-rozet I also want to clarify whether I understood correctly that the number of repositories now does not affect the rank in any way?

https://github.com/anuraghazra/github-readme-stats/blob/091e304e26a1562a58d6c2d05a066e29042b30a7/src/calculateRank.js#L84

The rest looks good to me. Great improvement of ranking system.

I also want you to know that since you became collaborator you can push your branches directly into this repository. It's better because Vercel test deployment workflow will work and will make testing and reviewing easier.

rickstaa:
> Don't have much idea about any of these ranking algorithm.
> 
> But whatever you folks think is a better system you can go ahead with but:
> 
> * Make sure to do a perf test once (basic test with [performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)API) to ensure API responses don't get effected since now we have those lookup arrays
> * Document the changes properly because external people might have no idea why their ranks changed and in that case you can refer them to the docs
> 
> @rickstaa @francois-rozet @qwerty541

@anuraghazra I didn't have time yet to review this, but maybe we should not add the QUANTILES themselves in the code 🤔. Sticking the weights and means already in the codebase is better.

https://github.com/anuraghazra/github-readme-stats/blob/24ac78bdb0725ebcb099cbea2fedbc258bdec15a/src/calculateRank.js#L35-L54

This prevents any performance problems from occurring, and if the code contains a comment that links to a `RANK_CALC.md` Markdown doc file in which the calculation and quantiles are explained, I think that is enough. What do you think? 

The ranking was improved since many users pointed out that it took a lot of work to get a rank other than A+ 😅. There were about three issues/discussions opened by people who were wondering what happened to their rank, but after I explained it to them and showed them that ranking up is now more accessible for the first ranks, no new issues were opened. However, we can document this change and the reasoning in a `RANK_CALC.md` doc file 👍🏻 .

francois-rozet:
To be sure, @rickstaa, you think that we should stick to the current system because it is simpler, even though it is a rough approximation?

rickstaa:
> To be sure, @rickstaa, you think that we should stick to the current system because it is simpler, even though it is a rough approximation of the true quantiles?

Hey @francois-rozet sorry for being unclear. My suggestion was that we can perform the calculation based on the quantiles offline and then just store the `mean` and `weights` in the code. This will prevent any performance issues and keeps the code clean. Then, We can store the calculation and quantiles somewhere and refer to them in a code comment 🤔.

francois-rozet:
You cannot get the true percentiles just based on the median. Currently, we assume that the statistics follow exponential/log-normal distributions, which implicitly fixes the quantiles with respect to the median. The percentiles we get are basically educated guesses.

If we want the true percentiles, we need the true quantiles.

qwerty541:
I have opened pull request #3141 which contains performance tests base implementation. After merging this one and pulling this branch with master we can check how quantiles affect on performance. But personally i do not expect much changes since there are no iterations through these quantiles arrays, it accessed everywhere by indexes.

francois-rozet:
Actually, there are iterations through the arrays with the `.findIndex()` calls. But this can be improved from $O(n)$ to $O(\log n)$ by using a bisection search. I can implement that.

qwerty541:
@francois-rozet Yes, you're right, i didn't noticed `.findIndex()` call on 2 line at first look.

https://github.com/anuraghazra/github-readme-stats/pull/2962/files#diff-d6a36eb9dcbdb8c6168a754251ba0c9318c7348ac6b17001cea3ba5aa84af728R2

I only noticed a call of the same function on line 136, but since this `THRESHOLDS` array contains only 9 elements and unlike from `score()` function it was called only once, I didn't attach much importance to it.

https://github.com/anuraghazra/github-readme-stats/pull/2962/files#diff-d6a36eb9dcbdb8c6168a754251ba0c9318c7348ac6b17001cea3ba5aa84af728R136

It would be great if you implement that.

I hope my code below will help.

```js
function findIndex(arr, target) {
  let left = 0;
  let right = arr.length - 1;
  let result = -1;

  while (left <= right) {
    const mid = Math.floor((left + right) / 2);

    if (arr[mid] < target) {
      result = mid; // Update the result and continue searching on the right half
      left = mid + 1;
    } else {
      right = mid - 1; // Search on the left half
    }
  }

  return result;
}

const sortedArray =  [
    0, 0, 0, 0, 2, 4, 8, 11, 15, 19, 23, 27, 32, 36, 41, 45, 50, 55, 60, 65, 71,
    76, 82, 87, 93, 99, 105, 111, 117, 124, 131, 137, 145, 151, 159, 166, 174,
    182, 190, 198, 207, 215, 225, 234, 244, 253, 264, 274, 285, 296, 306, 318,
    330, 342, 355, 368, 382, 396, 409, 424, 440, 457, 475, 493, 512, 531, 551,
    570, 593, 618, 643, 667, 695, 723, 752, 784, 815, 857, 893, 934, 984, 1037,
    1094, 1152, 1217, 1289, 1379, 1475, 1576, 1696, 1851, 2023, 2232, 2480,
    2835, 3242, 3885, 4868, 6614, 11801, 792319,
  ];
const targetValue = 2333;
const lowerIndex = findIndex(sortedArray, targetValue);

if (lowerIndex !== -1) {
  console.log(`Index of the first value lower than ${targetValue} is: ${lowerIndex}. The value is: ${sortedArray[lowerIndex]}`);
} else {
  console.log(`No value lower than ${targetValue} found.`);
}
```

rickstaa:
@francois-rozet, @anuraghazra just a small heads up that I just merged @qwerty541 [pull requests](https://github.com/anuraghazra/github-readme-stats/pull/3141) which allows us to safeguard against performance issues.

francois-rozet:
Hello @anuraghazra, @rickstaa, @qwerty541, I have (finally) replaced the linear-time search by a log-time bisection search. However, thinking back, I am now a bit skeptical about this PR. It would great if the percentiles were purely "data-driven", but I feel like it is not worth the increase of code complexity (and code (and data) maintenance). Users seem to be quite happy with the current ranking system, which is easy to understand and maintain, even if it is slightly off.

I believe #2637 is a much more important issue for the project.

BhasherBEL:
> @francois-rozet Yes, you're right, i didn't noticed `.findIndex()` call on 2 line at first look.
> 
> https://github.com/anuraghazra/github-readme-stats/pull/2962/files#diff-d6a36eb9dcbdb8c6168a754251ba0c9318c7348ac6b17001cea3ba5aa84af728R2
> 
> I only noticed a call of the same function on line 136, but since this `THRESHOLDS` array contains only 9 elements and unlike from `score()` function it was called only once, I didn't attach much importance to it.
> 
> https://github.com/anuraghazra/github-readme-stats/pull/2962/files#diff-d6a36eb9dcbdb8c6168a754251ba0c9318c7348ac6b17001cea3ba5aa84af728R136
> 
> It would be great if you implement that.
> 
> I hope my code below will help.
> 
> ```js
> function findIndex(arr, target) {
>   let left = 0;
>   let right = arr.length - 1;
>   let result = -1;
> 
>   while (left <= right) {
>     const mid = Math.floor((left + right) / 2);
> 
>     if (arr[mid] < target) {
>       result = mid; // Update the result and continue searching on the right half
>       left = mid + 1;
>     } else {
>       right = mid - 1; // Search on the left half
>     }
>   }
> 
>   return result;
> }
> 
> const sortedArray =  [
>     0, 0, 0, 0, 2, 4, 8, 11, 15, 19, 23, 27, 32, 36, 41, 45, 50, 55, 60, 65, 71,
>     76, 82, 87, 93, 99, 105, 111, 117, 124, 131, 137, 145, 151, 159, 166, 174,
>     182, 190, 198, 207, 215, 225, 234, 244, 253, 264, 274, 285, 296, 306, 318,
>     330, 342, 355, 368, 382, 396, 409, 424, 440, 457, 475, 493, 512, 531, 551,
>     570, 593, 618, 643, 667, 695, 723, 752, 784, 815, 857, 893, 934, 984, 1037,
>     1094, 1152, 1217, 1289, 1379, 1475, 1576, 1696, 1851, 2023, 2232, 2480,
>     2835, 3242, 3885, 4868, 6614, 11801, 792319,
>   ];
> const targetValue = 2333;
> const lowerIndex = findIndex(sortedArray, targetValue);
> 
> if (lowerIndex !== -1) {
>   console.log(`Index of the first value lower than ${targetValue} is: ${lowerIndex}. The value is: ${sortedArray[lowerIndex]}`);
> } else {
>   console.log(`No value lower than ${targetValue} found.`);
> }
> ```

As the size of the list is small and constant, it may be interesting to use a dictionary instead of a list and a "complex" find algorithm ?

src/calculateRank.js
```diff
@@ -1,12 +1,114 @@
-function exponential_cdf(x) {
-  return 1 - 2 ** -x;
+function searchSorted(arr, x) {
```
BhasherBEL, line 1: As the size of the list is small and constant, it may be interesting to use a dictionary instead of a list and a "complex" find algorithm ?

francois-rozet: I am not sure how you would use a dictionary to find the index `i` for which `arr[i - 1] <= x < arr[i]`.

PR #2963, opened 2023-07-22
Enable dependabot

qwerty541:


rickstaa (approved):

rickstaa:
@qwerty541 I just noticed that would cause a lot of monthly pull requests. Maybe we can ignore patch versions 🤔?

```yml
    ignore:
      - dependency-name: "*"
        update-types: ["version-update:semver-patch"]
```

qwerty541:
> @qwerty541 I just noticed that would cause a lot of monthly pull requests. Maybe we can ignore patch versions thinking?
> 
> ```yaml
>     ignore:
>       - dependency-name: "*"
>         update-types: ["version-update:semver-patch"]
> ```

Currently we got a lot if pull requests because we didn't maintain them for a long time. Let's check how it will going for 1-2 weeks and than make a decision.

rickstaa:
> > @qwerty541 I just noticed that would cause a lot of monthly pull requests. Maybe we can ignore patch versions thinking?
> > ```yaml
> >     ignore:
> >       - dependency-name: "*"
> >         update-types: ["version-update:semver-patch"]
> > ```
> 
> Currently we got a lot if pull requests because we didn't maintain dependencies for a long time. Let's check how it will going for 1-2 weeks and than make a decision.

Cool, I disable patches in most of my repositories but if it is manageable I'm okay with it 👍🏻.

PR #3028, opened 2023-08-01
Themes: Add test theme to check preview workflow

qwerty541:


rickstaa (approved):

rickstaa:
Were these themes generated using Co-Pilot 🤔? I like the calm_pink. They look different enough from the current themes. The only close relative is the `slateorange` and `holi`.

qwerty541 (commented):
> Were these themes generated using Co-Pilot thinking? I like the calm_pink. They look different enough from the current themes. The only close relative is the `slateorange` and `holi`.

@rickstaa Yes, you are right, this themes were generated using GitHub Copilot. I needed them to test if preview workflow works, but than i noticed that they looks nice and different from others, so i decided to keep this PR open.

rickstaa:
> > Were these themes generated using Co-Pilot thinking? I like the calm_pink. They look different enough from the current themes. The only close relative is the `slateorange` and `holi`.
> 
> @rickstaa Yes, you are right, this themes were generated using GitHub Copilot. I needed them to test if preview workflow works, but than i noticed that they looks nice and different from others, so i decided to keep this PR open.

Cool. Nice little experiment. Feel free to merge this pr if you want 👍.

PR #3042, opened 2023-08-05
Refactor: Top langs card: Resolve type error with hide_border variable

qwerty541:


qwerty541 (commented):
@rickstaa I see that you have doubts about the merging of this pull request, so I will try to explain the motivation for these changes. `hide_border` query option for top languages card is optional, so type of this variable is `boolean | undefined`. `setHideBorder()` function requires argument type `boolean`, so passing an argument with `boolean | undefined` type cause an error. To resolve this error i check if `hide_border` variable is defined before pass it inside `setHideBorder()` function.

rickstaa:
> @rickstaa I see that you have doubts about the merging of this pull request, so I will try to explain the motivation for these changes. `hide_border` query option for top languages card is optional, so type of this variable is `boolean | undefined`. `setHideBorder()` function requires argument type `boolean`, so passing an argument with `boolean | undefined` type cause an error. To resolve this error i check if `hide_border` variable is defined before pass it inside `setHideBorder()` function.

Ha, thanks for the explanation that makes the pull request a no-brainer 👍🏻. I understand the change but didn't yet find the time to review it. I just approved and merged it.

rickstaa (approved):

PR #3048, opened 2023-08-07
catppuccin-mocha theme added

wizsk:
It is a popular theme but it wasn't added here so I added it.


anuraghazra:
Already added in #2376

PR #3126, opened 2023-08-22
Top langs card: Remove unreachable code from fetcher and increase tests coverage

qwerty541:
There is two error handlers inside `fetchTopLanguages` function. Having error handler on strings 78-81 makes unreachable the next more advanced error handler on strings 84-102.


rickstaa (approved):

rickstaa:
good catch!

PR #3137, opened 2023-08-24
Enable curly eslint rule

qwerty541:
This pull request enables `curly` eslint rule.

https://eslint.org/docs/latest/rules/curly


rickstaa:
@qwerty541 although I'm okay with adding this rule since I use it myself, the choice to add this can be opinionated. I therefore asked @anuraghazra also to review this.

rickstaa (approved):

rickstaa:
I think we can safely add this when conflicts are resolved. Appart from your recent commit, the codebase doesn't contain code that violates this rule 👍🏻.

PR #3067, opened 2023-08-13
docs: fix relative README links

rickstaa:
Fixes #3066.

Issue #3066, opened 2023-08-13
README `src/calculateRank.js` Link Inconsistent


conorvenus:
### Describe the bug

 The [src/calculateRank.js](./src/calculateRank.js) link inside the README file does not work consistently, i.e., if you are on `https://github.com/anuraghazra/github-readme-stats` and you click it, it takes you to `https://github.com/anuraghazra/src/calculateRank.js` incorrectly.

### Expected behavior

I would expect to be taken to the `src/calculateRank.js` file within the current repository.

### Screenshots / Live demo link

https://streamable.com/ocjjn1

rickstaa:
Thanks for the bug report. This will be fixed when https://github.com/anuraghazra/github-readme-stats/pull/3067 is merged 👍🏻.

a-s-t-e-y-a:
@rickstaa I solved the readme.md issue by fixing the link to https://github.com/anuraghazra/github-readme-stats/blob/master/src/calculateRank.js

rickstaa:
Looks like there is a bug on GitHub's side. I just opened a ticket with the support.

![image](https://github.com/anuraghazra/github-readme-stats/assets/17570430/90c3a123-f6d0-4507-8a6b-6e63513cddcc)

> Originally posted in https://github.com/anuraghazra/github-readme-stats/pull/3140.

qwerty541:
> Looks like there is a bug on GitHub's side. I just opened a ticket with the support.
> 
> ![image](https://user-images.githubusercontent.com/17570430/274696333-90c3a123-f6d0-4507-8a6b-6e63513cddcc.png)
> 
> > Originally posted in https://github.com/anuraghazra/github-readme-stats/pull/3140.

Nice, let's wait for their response.

rickstaa:
> > Looks like there is a bug on GitHub's side. I just opened a ticket with the support.
> > ![image](https://user-images.githubusercontent.com/17570430/274696333-90c3a123-f6d0-4507-8a6b-6e63513cddcc.png)
> > > Originally posted in https://github.com/anuraghazra/github-readme-stats/pull/3140.
> 
> Nice, let's wait for their response.

I just got several responses:

![image](https://github.com/anuraghazra/github-readme-stats/assets/17570430/8179f51f-01bf-4857-b4e1-de4887cb7e39)

![image](https://github.com/anuraghazra/github-readme-stats/assets/17570430/d83308e2-d1db-481e-b244-9c2a0e2d2102)

[...]

PR #3140, opened 2023-08-25
Modified so that the `src/calculateRank.js` link can be used on the repository main screen

Yaminyam:
Modify the path to an absolute path so that you can use the link in other paths, such as the repository main screen.


qwerty541 (commented):
@Yaminyam Thanks for opening pull request!

@rickstaa I think we can merge this pull request as temporary solution for https://github.com/anuraghazra/github-readme-stats/issues/3066. Unfortunately this will not work in vscode locally and in browser when user on not master branch, but it is more important for this link to work on repository main screen. It is very rare situation when you need it somewhere else. What do you think?

rickstaa:
It looks like #3067 fixed all relative links except `vercel.json` and `src/calculaterank.js` 🤔. This is a bug since they should add the `blob/master` prefix. I will contact the support. If they can't fix it, let's use an absolute link. The benefit of a relative link is that it works on all forks and branches.

rickstaa:
@qwerty541, it just became a lot stranger. This bug is only present on GRS. See https://github.com/rickstaa/relative_md_links_file_bug?tab=readme-ov-file 🤔.

qwerty541:
> @qwerty541, it just became a lot stranger. This bug is only present on GRS. See https://github.com/rickstaa/relative_md_links_file_bug?tab=readme-ov-file 🤔. Let's see what the GitHub support says about this.

@rickstaa Yeah, thats strange, i just checked if it works in another repository with same paths https://github.com/qwerty541/relative_md_links_file_bug/tree/qwerty541-patch-1. We also can test if it works in master branch if you are up to merge.

rickstaa:
> > @qwerty541, it just became a lot stranger. This bug is only present on GRS. See https://github.com/rickstaa/relative_md_links_file_bug?tab=readme-ov-file 🤔. Let's see what the GitHub support says about this.
> 
> @rickstaa Yeah, thats strange, i just checked if it works in another repository with same paths https://github.com/qwerty541/relative_md_links_file_bug/tree/qwerty541-patch-1. We also can test if it works in master branch if you are up to merge.

Thanks for debugging! It appears to be a GitHub bug. See https://github.com/anuraghazra/github-readme-stats/issues/3066#issuecomment-1763309576.

qwerty541:
Closing this in favour of https://github.com/anuraghazra/github-readme-stats/pull/3389.

@Yaminyam Thanks for your efforts anyway.

PR #3145, opened 2023-08-25
added user profile option to rankIcon

### LINKED ISSUE #3128: Use profile picture as a rank icon
### Is your feature request related to a problem? Please describe.

I don't really want to show my rank on the stats page since its fairly low, but I think the GitHub icon could be improved by also allowing the use of the owners profile picture.

### Describe the solution you'd like

As the icon keyword you could use `profile` and then use the GitHub api to return the users profile picure.  
The same user account can be used as the given rank username for the vercel API.

### Describe alternatives you've considered

One alternative is allowing custom images but that would be more work than required for my request and would run into issues with pictures that dont conform to the circle used for the rank. So using the API is the simplier option especially since its already being used to retrieve stats for the rank.

### Additional context

_No response_

alex8obrien:
- Adds the option for a users GitHub profile picture to be used as a rank icon
- I'm unable to test if this works, please feel free to clone this and update me with the results

closes #3128

### DIFF
--- a/src/cards/stats-card.js
+++ b/src/cards/stats-card.js
@@ -394,7 +394,7 @@ const renderStatsCard = (stats, options = {}) => {
         <circle class="rank-circle-rim" cx="-10" cy="8" r="40" />
         <circle class="rank-circle" cx="-10" cy="8" r="40" />
         <g class="rank-text">
-          ${rankIcon(rank_icon, rank?.level, rank?.percentile)}
+          ${rankIcon(name, rank_icon, rank?.level, rank?.percentile)}
         </g>
       </g>`;
 
--- a/src/common/icons.js
+++ b/src/common/icons.js
@@ -17,13 +17,20 @@ const icons = {
 /**
  * Get rank icon
  *
+ * @param {string} username - The username for the GitHub Account
  * @param {string} rankIcon - The rank icon type.
  * @param {string} rankLevel - The rank level.
  * @param {number} percentile - The rank percentile.
  * @returns {string} - The SVG code of the rank icon
  */
-const rankIcon = (rankIcon, rankLevel, percentile) => {
+const rankIcon = (username, rankIcon, rankLevel, percentile) => {
   switch (rankIcon) {
+    case "profile":
+      return `
+        <svg>
+          <image href="https://avatars.githubusercontent.com/${username}" x="0" y="0" height="66" width="66" />
+        </svg>
+      `;
     case "github":
       return `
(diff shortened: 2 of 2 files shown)


src/common/icons.js
```diff
@@ -17,13 +17,20 @@ const icons = {
...
+ * @param {string} username - The username for the GitHub Account
  * @param {string} rankIcon - The rank icon type.
  * @param {string} rankLevel - The rank level.
  * @param {number} percentile - The rank percentile.
  * @returns {string} - The SVG code of the rank icon
  */
-const rankIcon = (rankIcon, rankLevel, percentile) => {
+const rankIcon = (username, rankIcon, rankLevel, percentile) => {
   switch (rankIcon) {
+    case "profile":
```
anuraghazra, line 28: Let's rename this to `avatar_image`

src/common/icons.js
```diff
@@ -17,13 +17,20 @@ const icons = {
...
  * Get rank icon
  *
+ * @param {string} username - The username for the GitHub Account
  * @param {string} rankIcon - The rank icon type.
  * @param {string} rankLevel - The rank level.
  * @param {number} percentile - The rank percentile.
  * @returns {string} - The SVG code of the rank icon
  */
-const rankIcon = (rankIcon, rankLevel, percentile) => {
+const rankIcon = (username, rankIcon, rankLevel, percentile) => {
```
anuraghazra, line 26: add the username param at the end instead of rearranging the params.

anuraghazra:
> I'm unable to test if this works, please feel free to clone this and update me with the results

Install the [vercel CLI ](https://vercel.com/docs/cli) and then just run `vercel dev` and open localhost:3000

src/cards/stats-card.js
```diff
@@ -394,7 +394,7 @@ const renderStatsCard = (stats, options = {}) => {
         <circle class="rank-circle-rim" cx="-10" cy="8" r="40" />
         <circle class="rank-circle" cx="-10" cy="8" r="40" />
         <g class="rank-text">
-          ${rankIcon(rank_icon, rank?.level, rank?.percentile)}
+          ${rankIcon(name, rank_icon, rank?.level, rank?.percentile)}
```
anuraghazra, line 397: The githubusercontent requires the username, eg "anuraghazra" it cannot work with the name eg "Anurag Hazra", so you need to pass down the username.

anuraghazra:
Seems like the alignment is needed 
image

rickstaa (changes requested):
@alex8obrien see requested changes of @anuraghazra.

PR #3156, opened 2023-08-27
Docs: Show additional stats icons

qwerty541:
At the moment these icons are not shown anywhere in documentation.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -577,7 +577,7 @@ Change the `?username=` value to your [Wakatime](https://wakatime.com) username.
 
 *   Showing additional stats
 
-![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=anuraghazra\&show=reviews,discussions_started,discussions_answered,prs_merged,prs_merged_percentage)
+![Anurag's GitHub stats](https://github-readme-stats.vercel.app/api?username=anuraghazra\&show_icons=true\&show=reviews,discussions_started,discussions_answered,prs_merged,prs_merged_percentage)
 
 *   Showing icons
 


rickstaa (approved):

rickstaa:
Hey @qwerty541, thanks for cleaning up this repository with all your PR 🚀.

One quick note. I see that you are using the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) convention, which is great! I however noticed that you capitalize both the label and the start of the commit while according to the convention these should be lowercase. Not a big deal but just something I noticed 👍🏻.

```text
docs: show additional stats icons (#3156)
```

vs 

```text
Docs: Show additional stats icons (#3156)
```

qwerty541:
> Hey @qwerty541, thanks for cleaning up this repository with all your PR 🚀.
> 
> One quick note. You are using the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) convention, which is excellent! I, however, noticed that you capitalize both the label and the start of the commit while, according to the convention, these should be lowercase. It's not a big deal, just something I noticed 👍🏻.
> 
> ```
> docs: show additional stats icons (#3156)
> ```
> 
> vs
> 
> ```
> Docs: Show additional stats icons (#3156)
> ```

To be honest, I have never read this convention. I'm just trying to use best practices from my experience for naming commits. It is also important to take into account that in different repositories the naming style of commits can change greatly due to some historical reasons, and when moving between projects, you need to change the naming style of commits. So when I first started contributing to this project, I adopted the commit naming style that was already used here. It's not difficult for me to write the label and commit message in lowercase if you think it's better that way. If we are going to use this https://www.conventionalcommits.org/en/v1.0.0/ convention, then it would be nice to add this requirement into our contributions guide https://github.com/anuraghazra/github-readme-stats/blob/master/CONTRIBUTING.md. I can create a pull request if you are up to approve it.

PR #3160, opened 2023-08-28
Update index.js

pranali18-Ai:
Changed the color of background


rickstaa:
Closed since I don't think we should change the default themes as we have a lot of available pieces, and people can customize the themes through API parameters in any way they like.. Feel free to explain why you think this is needed.

Issue #1935, opened 2022-08-10
Top Issues Dashboard


rickstaa:
A simple dashboard that lists the top issues/bugs/features and pull requests.

## Top issues

1. #1
2. #1471
3. #455
4. #1406
5. #1001

arndom:
@rickstaa can I be assigned this?

qwerty541:
> @rickstaa can I be assigned this?

Hey, @arndom! No, you can't be assigned. This is special automatically generated dashboard issue for tracking community demand. It should be never closed or assigned to someone for any reason. You can choose some issue that you are interested in working on from the list.

qwerty541:
> Oh ok, my bad. Maybe a label could be added to indicate that ?

This issue already has a special unique label "top issues dashboard".

arndom:
Ooh, didn't realize that, thanks.

FrankRay78:
Hello @rickstaa, I'm a maintainer over at [spectre.console](https://github.com/spectreconsole/spectre.console).  Limited maintainer capacity means we've been looking at ways to manage the volume of incoming issues/PRs. Your dashboard above looks most suitable. I don't (currently) know anything about github actions bots, but I'm wondering if it's something that could be copied to our repo.

rickstaa:
> Hello @rickstaa, I'm a maintainer over at [spectre.console](https://github.com/spectreconsole/spectre.console). Limited maintainer capacity means we've been looking at ways to manage the volume of incoming issues/PRs. Your dashboard above looks most suitable. I don't (currently) know anything about github actions bots, but I'm wondering if it's something that could be copied to our repo.
> 
> **Update** - I've just seen your repo: https://github.com/rickstaa/top-issues-action

Hey @FrankRay78, it's fantastic to hear that you like my action and are considering it for https://github.com/spectreconsole/spectre.console 🚀. The action targets open-source projects like GRS that don't have a clear backlog or development pipeline. If you have any questions or feature requests, don't hesitate to ask 🙂.

qwerty541:
> Hello @rickstaa, I'm a maintainer over at [spectre.console](https://github.com/spectreconsole/spectre.console). Limited maintainer capacity means we've been looking at ways to manage the volume of incoming issues/PRs. Your dashboard above looks most suitable. I don't (currently) know anything about github actions bots, but I'm wondering if it's something that could be copied to our repo.
> 
> **Update** - I've just seen your repo: https://github.com/rickstaa/top-issues-action

[...]

PR #3182, opened 2023-08-31
Feature team stats

pursonchen:
Hi, guys. This day I needed to evaluate a development team's GitHub status, and then I found your awesome tools, The card is very good, I made a little change to aggregate many accounts's statuses and then output a status card. the API is  https://host/api?team=account1,account2

### DIFF
--- a/api/team.js
+++ b/api/team.js
@@ -39,12 +39,13 @@ export default async (req, res) => {
     border_color,
     rank_icon,
     show,
+    team_name,
   } = req.query;
   res.setHeader("Content-Type", "image/svg+xml");
 
   const username_list = usernames.split(",");
 
-  for (const username in username_list) {
+  for (const username of username_list) {
     if (blacklist.includes(username)) {
       return res.send(renderError("Something went wrong"));
     }
@@ -65,11 +66,11 @@ export default async (req, res) => {
     totalDiscussionsStarted: 0,
     totalDiscussionsAnswered: 0,
     contributedTo: 0,
-    rank: { level: "C", percentile: 100 },
+    rank: { level: "C", percentile: 0 },
   };
 
   try {
-    for (let username in username_list) {
+    for (let username of username_list) {
       const stats = await fetchStats(
         username,
         parseBoolean(include_all_commits),
@@ -87,7 +88,8 @@ export default async (req, res) => {
       team_stats.contributedTo += stats.contributedTo;
       team_stats.rank.percentile += stats.rank.percentile;
     }
-    team_stats.name = `${username_list}'s team `;
+    team_stats.name = `${team_name ? team_name : username_list[0]} team `;
+
     team_stats.rank.level = rank_to_level(
       team_stats.rank.percentile / 100 / username_list.length,
     );
--- a/api/team.js
+++ b/api/team.js
@@ -1,0 +1,142 @@
+import { renderStatsCard } from "../src/cards/stats-card.js";
+import { blacklist } from "../src/common/blacklist.js";
+import {
+  clampValue,
+  CONSTANTS,
+  parseArray,
+  parseBoolean,
+  renderError,
+} from "../src/common/utils.js";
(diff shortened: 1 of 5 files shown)


rickstaa:
@pursonchen Although I love the feature you have implemented, we can, in the current state, not merge it into the master branch. This is because it performs N GraphQL requests, causing our already overloaded public instance rate-limited faster (i.e. https://docs.github.com/en/graphql/overview/resource-limitations) 😅. See https://github.com/anuraghazra/github-readme-stats/issues/1471 for more information. I would, however, like to add your card under an environmental variable when we release our GitHub action later this year (i.e. https://github.com/anuraghazra/github-readme-stats/issues/2179). Could you [create a feature request](https://github.com/anuraghazra/github-readme-stats/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.yml) so we can track your feature and see [how much interest there is from the community](https://github.com/anuraghazra/github-readme-stats/issues/1935)?

api/team.js
```diff
@@ -0,0 +1,144 @@
...
+      team_stats.mergedPRsPercentage += stats.mergedPRsPercentage;
+      team_stats.totalReviews += stats.totalReviews;
+      team_stats.totalCommits += stats.totalCommits;
+      team_stats.totalIssues += stats.totalIssues;
+      team_stats.totalStars += stats.totalStars;
+      team_stats.totalDiscussionsStarted += stats.totalDiscussionsStarted;
+      team_stats.totalDiscussionsAnswered += stats.totalDiscussionsAnswered;
+      team_stats.contributedTo += stats.contributedTo;
+      team_stats.rank.percentile += stats.rank.percentile;
+    }
```
qwerty541, line 90: @pursonchen I think that the following lines of code should be totally reworked if we are going to merge this feature someday. I expect to have only one GraphQL request which will fetch required data about team members. Calling `fetchStats()` function inside loop is an error since it decrease performace and cause hitting upstream GitHub API rate limits. Algorithm complexity should be improved from $O(n)$ to $O(1)$ as previously noticed @rickstaa.

I also do not see much sense in showing sum stats of n people and their average rank. Is showing rank useful in case of team card? I do not think so. I think it is a place for discussion.

Remember that you can implement anything and deploy it on your own Vercel instance.

rickstaa: @qwerty541, excellent point! It seems that with [GraphQL v4](https://docs.github.com/en/graphql), data for multiple users can now be fetched without using more GraphQL points:

```graphql
{
  user1: user(login: "rickstaa") {
    ...UserFragment
  }
  user2: user(login: "qwerty541") {
    ...UserFragment
  }
  user3: user(login: "pursonchen") {
    ...UserFragment
  }
  rateLimit {
    limit
    cost
    remaining
    resetAt
  }
}

fragment UserFragment on User {
  login
  name
  contributionsCollection {
    totalCommitContributions
    totalPullRequestReviewContributions
  }
  repositoriesContributedTo(
    first: 1
    contributionTypes: [COMMIT, ISSUE, PULL_REQUEST, REPOSITORY]
  ) {
    totalCount
  }
  pullRequests(first: 1) {
    totalCount
  }
  mergedPullRequests: pullRequests(states: MERGED) {
    totalCount
  }
  openIssues: issues(states: OPEN) {
    totalCount
  }
  closedIssues: issues(states: CLOSED) {
    totalCount
  }
  followers {
    totalCount
  }
  repositoryDiscussions {
    totalCount
  }
  repositoryDiscussionComments(onlyAnswers: true) {
    totalCount
  }
}
```

I was surprised to learn about that feature and that it only uses 1 GraphQL point. Thanks for bringing it to my attention. As a result, making the addition of this card possible without having to wait for the GitHub action to be released. 

However, creating a separate team data fetcher is essential so that the codebase doesn't get cluttered. Furthermore, we have to discuss which stats would be interesting to show for teams. I think the following stats are not helpful when looking at teams:

- Contributed to (last year).
- Total discussions started.
- Total discussion answered.
- PR reviews.

PR #3186, opened 2023-09-01
tests(stats card fetcher): add upstream API error test when include_all_commits true

qwerty541:
This pull request covers with tests these lines of code:

https://app.codecov.io/gh/anuraghazra/github-readme-stats/blob/master/src%2Ffetchers%2Fstats-fetcher.js#L178

### DIFF
--- a/tests/fetchStats.test.js
+++ b/tests/fetchStats.test.js
@@ -237,6 +237,37 @@ describe("Test fetchStats", () => {
     });
   });
 
+  it("should return 0 commits when all_commits true and API returns error", async () => {
+    mock
+      .onGet("https://api.github.com/search/commits?q=author:anuraghazra")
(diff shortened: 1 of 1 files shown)


rickstaa (approved):

qwerty541 (commented):
> @ qwerty541 I approved this pull request since it improves the test coverage. However, it is an excellent opportunity to discuss if the current behavoir is desirable. Would it, for example, be better to return an error and let the card retry later instead of showing `0` commits? I have had multiple issues open about people complaining that `0` commits were shown. I think #2448 will improve this behavoir already, but I was curious to to know what you think.

@rickstaa I have opened a new issue with my thoughts about this behavior https://github.com/anuraghazra/github-readme-stats/issues/3225. Let's merge this pull request for now to increase tests coverage.

PR #3191, opened 2023-09-02
refactor: fix card colors type and function to resolve vscode type errors

qwerty541:
Current jsdoc is wrong since only card background color can be an array(gradient).


rickstaa (approved):

rickstaa:
Good catch!

PR #3216, opened 2023-09-10
refactor: use more clear retryer error messages

qwerty541:


rickstaa (approved):

src/common/retryer.js
```diff
@@ -1,10 +1,12 @@
 import { CustomError, logger } from "./utils.js";
 
 // Script variables.
+
+// Count the number of GitHub API tokens available.
 const PATs = Object.keys(process.env).filter((key) =>
   /PAT_\d*$/.exec(key),
 ).length;
-const RETRIES = PATs ? PATs : 7;
+const RETRIES = process.env.NODE_ENV === "test" ? 7 : PATs;
```
rickstaa, line 9: @qwerty541 Although this PR is a great addition, I think the process could have been better.NODE_ENV variable is needed. The vercel variables should be available locally when running `vercel dev` (see https://vercel.com/docs/projects/environment-variables#environments).

rickstaa (changes requested):

src/common/retryer.js
qwerty541: @rickstaa I did it this way since this code was already used in another parts of projects, for example https://github.com/anuraghazra/github-readme-stats/blob/master/src/getStyles.js#L134. As i understand `NODE_ENV` is not vercel variable, it was not set on production or local development, it was set by jest framework only for testing purposes, see https://jestjs.io/docs/environment-variables#node_env

rickstaa: Cool, let's merge this then 👍🏻.

rickstaa (approved):

PR #3232, opened 2023-09-15
refactor: resolve vscode type errors in wakatime card render and remove redundant css

qwerty541:


rickstaa:
@qwerty541, thanks for the pull request. Here, I would do one of the following:

1. Remove the global `getStyles` function and replace it with a `getRankCardStyle` and `getWakatimeCardStyle` function inside the wakatime and stats card files.
2. OR: modify the global `getStyles` function to have a `card_type` variable.

Your current PR causes two `getStyles` functions to exist in the codebase, which is not bad but might confuse people. I prefer option 1, but since @anuraghazra initially added this function, he may give his thoughts before we merge this 👍🏻.

qwerty541 (commented):
> @qwerty541, thanks for the pull request. Here, I would do one of the following:
> 
> 1. Remove the global `getStyles` function and replace it with a `getRankCardStyle` and `getWakatimeCardStyle` function inside the wakatime and stats card files.
> 2. OR: modify the global `getStyles` function to have a `card_type` variable.
> 
> Your current PR causes two `getStyles` functions to exist in the codebase, which is not bad but might confuse people. I prefer option 1, but since @anuraghazra initially added this function, he may give his thoughts before we merge this 👍🏻.

@rickstaa Thanks for review, i agree with your thoughts about removing global `getStyles` function. I have updated the code, check it out.

src/cards/stats-card.js
```diff
@@ -76,6 +75,118 @@ const createTextNode = ({
...
+      }
+      to {
+        stroke-dashoffset: ${calculateCircleProgress(progress)};
+      }
+    }
+  `;
+};
+
+/**
+ * Retrieves CSS styles for a card.
```
rickstaa, line 119:
```suggestion
 * Retrieves CSS styles for the card.
```

src/cards/wakatime-card.js
```diff
@@ -158,6 +157,36 @@ const recalculatePercentages = (languages) => {
   });
 };
 
+/**
+ * Retrieves CSS styles for a card.
```
rickstaa, line 161:
```suggestion
 * Retrieves CSS styles for the card.
```

rickstaa (approved):

PR #3283, opened 2023-09-24
infra: enable no-negated-condition eslint rule

qwerty541:
https://eslint.org/docs/latest/rules/no-negated-condition


rickstaa:
@qwerty541, why do you want to turn this rule off?

qwerty541 (commented):
> @qwerty541, why do you want to turn this rule off?

@rickstaa I made a mistake, i want to enable this rule, thanks for catch. I have update pull request.

rickstaa (approved):

Issue #3184, opened 2023-08-31
add border for "graphs"


Arbee4ever:
### Is your feature request related to a problem? Please describe.

Some Languages colors blend into the background of my "top languages" card ![image](https://github.com/anuraghazra/github-readme-stats/assets/45911516/0603f76c-e27c-44fc-a6fc-485a4eb799c1)

### Describe the solution you'd like

I would like to have a white or some other color border around stuff like the top languages bar, to increase its visibility and looks.

ShambhavII-Singh:
Hi can you assign this issue to me? I would like to work on this.

rickstaa:
> Hi can you assign this issue to me? I would like to work on this.

Yea of course! Thanks for considering contributing to our project 🎉. I'm looking forward to your pull request.

aialok:
Can I work on this project ?

vinfinity7:
HI , I seem to be working on the issues but am not able to find where and how to review and test the changes that I am doing , Kindly assist 
@rickstaa

Pratik1603:
Please assign this to me

rickstaa:
@aialok & @Pratik1603, this issue was assigned to @ShambhavII-Singh, but I have not yet received a PR. It looks like @vinfinity7 is also working on this issue.

qwerty541:
@aialok @Pratik1603 I think one of you can work on this issue if still want. Since we didn't received a pull request from @ShambhavII-Singh and pull request submitted by @vinfinity7 is incorrect. Please read my review https://github.com/anuraghazra/github-readme-stats/pull/3282#pullrequestreview-1641076786 under declined pull request before you start the work.

aialok:
I will be working on this issue. 
will report you soon and make a pull request.

qwerty541:
> I will be working on this issue. will report you soon and make a pull request.

@aialok Thanks for considering contributing to our project. I have assigned you to this issue. I'm looking forward to your pull request.

aialok:
@qwerty541  I am trying to run the server on localhost but it redirect me to the main URL. I have tried to fix the problem but I am not able to see the changes. Can you tell me how to setup localhost ? I tried setting it up from contributing[dot].md but it did'nt seems like working.

Arbee4ever:
[...]

PR #3314, opened 2023-10-02
Fix/add border to graphs 3184

imrushikesh77:
Added white stroke to graphs in top languages card


rickstaa:
@imrushikesh77, thanks for your pull request. I, however, think this was different from what was meant in https://github.com/anuraghazra/github-readme-stats/issues/3184. https://github.com/anuraghazra/github-readme-stats/issues/3184 is about having the ability to add a border colour around the language percentage display element.

PR #3319, opened 2023-10-03
Algeria flag theme with light and dark options (algeria_flag_light, algeria_flag_dark)

iaminebriki:
I've added Algeria flag theme (name: algeria_flag):

preview:
algeria_flag


anuraghazra:
@iaminebriki I think it's better if we just keep 1 version of the flag's color. 
We should retain the original color of the flag instead of modifying it with a dark variant.

iaminebriki:
> @iaminebriki I think it's better if we just keep 1 version of the flag's color. We should retain the original color of the flag instead of modifying it with a dark variant.

@anuraghazra okey sure!

iaminebriki:
@anuraghazra please check #3420 (updated PR)

Issue #1351, opened 2021-10-03
Add Indian language doc translation


susheelg1197:
Hi, kindly assign me this issue. I will translate the repository into "HINDI" language.

rickstaa:
Related to #1349.

@susheelg1197 Thanks for your issue. Feel free to create a pull request.

yavinash249:
Please assign me this issue. I'll do it in two languages.

qwerty541:
Hey, @yavinash249! Thanks for your interest in our project! We are going to remove docs translations at all in future because we are unable to properly maintain them. Most of docs translations currently outdated and confuse users. I think this issue can be closed.

PR #3339, opened 2023-10-09
#1351 done translate in hindi

yavinash249:
translate readme in hindi language ,  not all term are translate but gernal term translate for user


rickstaa:

PR #3347, opened 2023-10-12
docs: improve vercel badge position

rickstaa:
This pull request improved the Vercel badge position since they have been a massive sponsor since GRS began.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,6 +24,9 @@
     </a>
     <br />
     <br />
+    <a href="https://vercel.com?utm\_source=github\_readme\_stats\_team\&utm\_campaign=oss">
+      <img src="./powered-by-vercel.svg"/>
+    </a>
   </p>
 
   <p align="center">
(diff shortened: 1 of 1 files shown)


rickstaa:
@qwerty541 On second though, I think we can have the badge both at the top and the bottom 👍🏻.

qwerty541 (approved):
@rickstaa I'm okay with having Vercel badge at both positions. Since i do not know much they sponsored this project, it's hard to judge for me if we need two badge or it's enough to have just one. Let's give @anuraghazra ability to make a decision about this. Judging by the request parameters in the link, Vercel gives the project a percentage for each user who followed the link and spent money. If this is true, then of course it is worth making the link more visible.

rickstaa:
> @rickstaa I'm okay with having Vercel badge at both positions. Since i do not know how much they sponsored this project, it's hard to judge for me if we need two badge or it's enough to have just one. Let's give @anuraghazra ability to make a decision about this. Judging by the request parameters in the link, Vercel gives the project a percentage for each user who followed the link and spent money. If this is true, then of course it is worth making the link more visible.

@qwerty541 as far as I understand it Vercel donated the bandwidth to keep the public Vercel instance running by providing @anuraghazra with a free enterprise subscription (see https://vercel.com/pricing). @anuraghazra feel free to add to this 👍🏻.

PR #3358, opened 2023-10-14
docs(contributing): remove outdated local dev steps

rickstaa:
This pull request removes some local development steps found in the contributing guide that are no longer required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,11 +34,8 @@ _(make sure you already have a [Vercel](https://vercel.com/) account)_
 2.  Fork the repository and clone the code to your local machine.
 3.  Run `npm install` in the repository root.
 4.  Run the command `vercel` in the root and follow the steps there.
-5.  Open `vercel.json` and set the maxDuration to 10.
-6.  Create a `.env` file in the root of the directory.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,11 +34,8 @@ _(make sure you already have a [Vercel](https://vercel.com/) account)_
 2.  Fork the repository and clone the code to your local machine.
 3.  Run `npm install` in the repository root.
 4.  Run the command `vercel` in the root and follow the steps there.
-5.  Open `vercel.json` and set the maxDuration to 10.
-6.  Create a `.env` file in the root of the directory.
-7.  In the .env file add a new variable named `PAT_1` with your [GitHub Personal Access Token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token).
```
qwerty541, line 39: @rickstaa I was very surpsired that after removing `PAT_1` variable from `.env` file local development still works because i do not understand why. I expected that endpoint should return error card because of the following strings: 

https://github.com/anuraghazra/github-readme-stats/blob/master/src/common/retryer.js#L25-L27

Can it be that PAT was cached somewhere and still required for the first run?

rickstaa: It supprised me as well when I found out. As can be seen from the documentation Vercel automatically downloads the environment variables into memory when running the `vercel dev` command (see https://vercel.com/docs/projects/environment-variables#development-environment-variables).

qwerty541 (approved):

Issue #2053, opened 2022-09-25
Migrate to automatic README translation


rickstaa:
**Is your feature request related to a problem? Please describe.**

Keeping the documentation up to date and managing the PRs would be more manageable if we switched from manual to automatic README translations (see https://github.com/dephraiim/translate-readme). The downside is that there might be some errors, but this shouldn't matter for understanding how to use GRS. Google has become quite good at translating languages in the last few years. The upside is that we no longer need to look at translation PRs, we can support more language, and the translations are up to date. We can add flags to the readme for people to choose their language.

Pranav2612000:
Hey @rickstaa 
I would like to work on this one. Is it fine?

rickstaa:
@Pranav2612000 First of all, welcome to the commuExcellent! Amazing that you want to help us improve the maintainability of the repository. I am unsure how hard it is to implement this feature and whether https://github.com/dephraiim/translate-readme serves our needs. The implementation found in translate-readme is quite basic (see https://github.com/dephraiim/translate-readme/blob/main/index.js). It could therefore be that it does not filter the query parameters found in the code blocks.

![image](https://user-images.githubusercontent.com/17570430/193402202-90dc02b3-ae31-45c7-a93b-7ebac626658d.png)

It could therefore be that we need to improve this action or create a new action. 

### My original idea

- I wanted to eliminate the manual translations that users add since these are outdated and hard to maintain.
- I wanted a GitHub action that translates the README into other languages and adds these translations to the repository.
- This action should keep code blocks that contain the query parameters in English.
- Ideally, this action should add the available translations to the main README. However, this can also be done manually. I was planning to add these using flags on mac and Linux and country codes on windows (see https://github.com/missive/emoji-mart/issues/559).

<div align="center">🇳🇱 🇫🇷 🇺🇸 🇩🇪 🇮🇹 </div>

rickstaa:
If you want, you can run some initial tests to see how good the https://github.com/dephraiim/translate-readme is and whether it is sufficient. Feel free to tag me if you need any help or have questions about the implementation 👍🏻.

rickstaa:
[...]

PR #3360, opened 2023-10-14
docs: add translations contribution guide

qwerty541:


rickstaa:
@qwerty541 I agree that something like this should be in the CONTRIBUTING.md file 👍🏻. One improvement to the translation pipeline would be to use Crowdin for both the cards and README translations (#2053 and #3364).

rickstaa (approved):

PR #3372, opened 2023-10-16
tests: add gist card performance test

qwerty541:


tests/bench/gist.bench.js
```diff
@@ -0,0 +1,51 @@
...
+            size: 85858,
+          },
+        ],
+      },
+    },
+  },
+};
+
+const mock = new MockAdapter(axios);
+mock.onPost("https://api.github.com/graphql").reply(200, gist_data);
```
anuraghazra, line 35: FYI. if we are mocking the API response, it's almost useless to do any performance testing. Since majority of performance bottlenecks are from the API only. 

Concatenating strings is hardly the performance intensive task.

qwerty541: Yes, I know. However, the work of generating a card takes as much as 100ms, and this test will allow us to prevent this indicator from increasing by tracking it.

Making a good test to measure the performance of real APIs is quite difficult since they will constantly fail with third-party errors, for example because of hitting rate limit or etc. Also the time of such test will always differ, not depending from changes in the code, but depending from load on GitHub servers. I'm not sure if it even worth to create such test.

rickstaa: @anuraghazra Although I agree with you that most of the performance bottlenecks are on the upstream API's side, I think 
having a performance test in the repository could give us some insight into whether the performance is decreased when implementing big changes.

For example, we added the stats card performance test to see if adding a data-based ranking mechanism would decrease the performance too much (i.e. https://github.com/anuraghazra/github-readme-stats/pull/2962). We can, of course also keep the stats card performance test and omit the test for the other cards, but I don't think adding them has significant downsides 🤔.

qwerty541: Maybe i'm wrong about that it worthless to create performance tests which works with real API. We already have e2e tests which works pretty good, almost never don not fail with third-party errors. Also affecting of the load on GitHub server to performance tests results will be reduced by multiple running and showing average time. I'm okay to rework these tests with real API if you gyus think that it will be much better.

rickstaa: > Maybe i'm wrong about that it worthless to create performance tests which works with real API. We already have e2e tests which works pretty good, almost never don not fail with third-party errors. Also affecting of the load on GitHub server to performance tests results will be reduced by multiple running and showing average time. I'm okay to rework these tests with real API if you gyus think that it will be much better.

In my opinion, the performance tests should be about the code performance. In contrast, the e2e tests should analyse the entire performance and notify use if something goes wrong when using the actual API. I, however, might be biased since I created the e2e tests 😅. Let's wait on @anuraghazra's opinion 👍🏻.

rickstaa (approved):

tests/bench/gist.bench.js
qwerty541: @rickstaa What do you think about merging this pull request and https://github.com/anuraghazra/github-readme-stats/pull/3374? We already agreed that it can be useful when implementaing significant changes and we didn't received response from @anuraghazra for a month. Anyway we can rework or remove it anytime.

rickstaa: Feel free to merge it 👍🏻. Looks like a useful addition to me.

PR #3398, opened 2023-10-20
strawberry inspired theme

uma-e:
added a theme inspired by strawberries that I initially created to customize my own stats card


anuraghazra:
@uma-e For themes which are custom and for your own use please consider using them via color codes instead of adding them to the github-readme-stats's theme.js.

PR #3403, opened 2023-10-21
Update translations.js

IORD1:
added a translation for 'archived' in Marathi as mr : 'संग्रहित' , in translation.js


anuraghazra:
When adding new locale in translation.js please do translation for the every cards, adding it particularly for 1 single field will cause runtime errors.

PR #3407, opened 2023-10-22
docs(theme): auto update theme readme

qwerty541:


rickstaa:
@qwerty541, why is this pull request required? That file should be automatically created using the https://github.com/anuraghazra/github-readme-stats/blob/master/.github/workflows/generate-theme-doc.yml action 🤔.

qwerty541 (commented):
> @qwerty541, why is this pull request required? That file should be automatically created using the https://github.com/anuraghazra/github-readme-stats/blob/master/.github/workflows/generate-theme-doc.yml action 🤔.

I think we always did themes readme update like this, `generate-theme-doc` workflow only regenerate readme and push changes into `updated-theme-readme` branch, it does not automatically create a pull request or etc. Of course these behavior can be improved but at the moment we have what we have.

rickstaa:
> > @qwerty541, why is this pull request required? That file should be automatically created using the https://github.com/anuraghazra/github-readme-stats/blob/master/.github/workflows/generate-theme-doc.yml action 🤔.
> 
> I think we always did themes readme update like this, `generate-theme-doc` workflow only regenerate readme and push changes into `updated-theme-readme` branch, it does not automatically create a pull request or etc. Of course these behavior can be improved but at the moment we have what we have.

Mm, I thought I improved that behaviour 🤔 . But I see you are right.

rickstaa (approved):

PR #3421, opened 2023-10-24
[Feature] Update repo-card.js #2951

Simer13:
Made some minor changes to enhance the repo-card, in order to enhance the presentation of the repo and make it look better.

### DIFF
--- a/src/cards/repo-card.js
+++ b/src/cards/repo-card.js
@@ -15,7 +15,12 @@ import {
 } from "../common/utils.js";
 import { repoCardLocales } from "../translations.js";
 
-const ICON_SIZE = 16;
+const ICON_SIZE = 20;
+const badge_text_color = "#FFFFFF";
+const badge_background_color = "E44D26";
+const primary_text_color = "4A90E2";
+const badge_text_fill_opacity=0.8;
+
 
 /**
  * Retrieves the repository description and wraps it to fit the card width.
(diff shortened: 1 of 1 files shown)


qwerty541 (commented):
Hey, @Simer13! I see two problems with this pull request.

First, the default behavior of the card should be kept as is, our users do not expect that cards in their profiles will be suddenly changed without action from their side. If you want repository card to have an alternate layout, it should be implemented as querystring parameter, same way as it was already implemented with our top languages card https://github.com/anuraghazra/github-readme-stats#demo-2.

And the second, if you want to add this for your personal use, you are able to deploy your branch on [own Vercel instance](https://github.com/anuraghazra/github-readme-stats#on-vercel), let's first create feature request issue and track if our community want it to be implemented since there is no sense in adding this if no one will use it.

Let's also wait for other maintainers input.

Simer13:
@qwerty541 Thanks for the input. You are right we can implement this as a new feature rather than making changes to the original card itself. We can create a new issue in relation to this topic and wait for the community's input and interest in the suggested feature. The new feature can be given either as an option or alternative to the user to select according to their type. 
Furthermore, we can also create a feature in another issue to let them modify the card in the way they want to represent it for their particular instance.

qwerty541:
> @qwerty541 Thanks for the input. You are right we can implement this as a new feature rather than making changes to the original card itself. We can create a new issue in relation to this topic and wait for the community's input and interest in the suggested feature. The new feature can be given either as an option or alternative to the user to select according to their type. Furthermore, we can also create a feature in another issue to let them modify the card in the way they want to represent it for their particular instance.

@Simer13 Let's start from creating [feature request issue](https://github.com/anuraghazra/github-readme-stats/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.yml) and wait for community's response in form of 👍. Please attach there a screenshot of the new card appereance.

PR #3431, opened 2023-10-27
Fix typo in top-languages-fetcher.js

eternityduck:


qwerty541 (approved):
Hey, @eternityduck! Thanks for you pull request! Good catch! Just one moment is wrong that the error message also should be changed inside tests, it is the reason why tests CI workflow failed.

PR #3432, opened 2023-10-27
chore: refactor isLocaleAvailable, move variable check to the method

eternityduck:
I have refactored the boolean check of variable, so the code will be prettier with only one check inside method instead of anywhere checking it. Approve if you find it useful

### DIFF
--- a/api/gist.js
+++ b/api/gist.js
@@ -26,7 +26,7 @@ export default async (req, res) => {
 
   res.setHeader("Content-Type", "image/svg+xml");
 
-  if (locale && !isLocaleAvailable(locale)) {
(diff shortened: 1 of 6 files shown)


qwerty541 (commented):
Hey, @eternityduck! Thanks for your pull request! In my opinion code will be more clear and readable when one function perform only one task. Currently `isLocaleAvailable` function just check if provided locale is available, but with this changes it also starts to validate input argument. Honestly, i do not like the change, but let's also wait for other maintainers thoughts about it. Also note that tests workflow failed, it measns that this changes breaks some expected behavior.

PR #3453, opened 2023-11-02
feature(repo card): add description lines count query parameter

qwerty541:
Before:

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>

After (`&description_lines_count=3`):

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>

After (`&description_lines_count=1`):

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>


readme.md
```diff
@@ -386,6 +386,7 @@ If we don't support your language, please consider contributing! You can find mo
 #### Repo Card Exclusive Options
 
 *   `show_owner` - Shows the repo's owner name *(boolean)*. Default: `false`.
+*   `description_lines_count` - Manually set the number of lines for the description *(number)*. If this parameter is not specified, the number of lines will be automatically adjusted according to the length of the description.
```
rickstaa, line 389: Maybe we can state that the maximum is 3.

rickstaa: Further, it looks like when the description is longer than the set `description_lines_count`, the maximum is used. You can check this out by looking at https://github.com/doersino/tweakpane-plugin-infodump. Here, we can do two things, or we use an ellipsis and truncate the text or document the behavoir.

src/cards/repo-card.js
```diff
@@ -81,8 +85,18 @@ const renderRepoCard = (repo, options = {}) => {
...
-  const descriptionLines = multiLineDescription.length;
+  const multiLineDescription = wrapTextMultiline(
+    desc,
+    DESCRIPTION_LINE_WIDTH,
+    DESCRIPTION_MAX_LINES,
+  );
+  const descriptionLines = description_lines_count
+    ? clampValue(
+        description_lines_count,
+        multiLineDescription.length,
```
rickstaa, line 96: I think this should be equal to the minimum required length (i.e. `1`).

```suggestion
        1,
```

rickstaa: I think the logic can be improved. I think the desired behavior would be:

- If the user sets 1 and the description is 3 lines, it will be truncated to 1 line.
- If the user doesn't specify anything, the length of multiLineDescription will be used.
- If the user sets 4 while 3 is the maximum allowed, it will be truncated to 3 lines.
- If the user sets 3 while the description is only 1 line, three lines will be utilized.

rickstaa (changes requested):
Thanks for your pull request. I added some comments 👍🏻.

src/cards/repo-card.js
qwerty541: Current implementation match second, third and forth items from your list. About the first item, i have doubts that we need to truncate the repository description if lines count provided by user is lower than actual.

I can't imagine why someone wants to show just part of his repository description. The only objective of this feature is giving ability to make size of different repositories cards equal, so if you have repository with long three lines description and the second one with short one line description, then you're going to extend the short one with empty lines, not vice versa. So i intentionally made clamping between actual (1-3) and maximum (3) lines count.

If we are going to keep current behavior i agree that it should be documented more properly.

rickstaa: While I understand that the primary use case for this feature is to ensure uniformity in the size of repository cards, it's worth considering that some users might prefer creating smaller cards with just a single description line. My suggestion for point 1 stems from the idea that this behaviour aligns more intuitively with users' expectations, eliminating the need for extensive documentation to explain its reasoning.

qwerty541: Ok, i agree with your approach, i have update the code and attached some examples of current behavior below, check it please.

Before:

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>

After (`&description_lines_count=3`):

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>

After (`&description_lines_count=1`):

<div align="center"> 
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
<a href="https://github.com/anuraghazra/github-readme-stats">

</a>
</div>

rickstaa (approved):
Approved. I asked @anuraghazra for a last review, but if documentation and tests are added, I now think this feature is good to merge.

qwerty541:
> Approved. I asked @anuraghazra for a last review, but if documentation and tests are added, I now think this feature is good to merge.

I added test in my recent commit, please check it. And this pull request already contains documentation.

qwerty541:
Or do you mean adding examples using this feature?

rickstaa:
> Or do you mean adding examples using this feature?

No, the pull request now looks good! I approved it so that it can be merged whenever you're ready 👍🏻.

PR #3457, opened 2023-11-05
infra: enable constructor-super eslint rule

qwerty541:
https://eslint.org/docs/latest/rules/constructor-super


rickstaa (approved):
@qwerty541, looks good from my end. If @anuraghazra wants to revert an ESLint rule, he can always do so.

PR #3496, opened 2023-11-22
feat(wakatime card): add disable animations query option

qwerty541:


rickstaa (approved):
Looks good to me. Good catch! I was about to say that it could use some tests, but I see those are also not present for the other cards, so feel free to merge it like this 👍🏻.

PR #3521, opened 2023-12-03
ci: use same .yml file extension for all workflows for consistency

qwerty541:


rickstaa (approved):
Feel free to merge this. I think the [general recommendation](https://yaml.org/faq.html) is to use `yaml` but GitHub seems to use `yml` in their documentation (see https://docs.github.com/en/actions/using-workflows/about-workflows#create-an-example-workflow).

PR #3559, opened 2023-12-24
Update readme.md

suraweera-AGSS:


touero (commented):
This is not necessary.

qwerty541 (commented):
Definitely agree with @touero. I don't see any sense in this change, so I'm closing this pull request.

PR #3577, opened 2024-01-03
Add the possibility of hosting the API on docker

Nerexbcd:
In my case I have a HomeLab at my house and I mainly use docker to selfhost my services, and thinking there might be more people like me, I created the possibility of selfhosting it with docker.


JPFrancoia:
This is an excellent idea, thank you.

PR #3644, opened 2024-02-11
fix: default value of cache_seconds in README

valbeat:
Hi Team!
I fixed the description in the README as it differs from the actual code.


qwerty541 (approved):
Good catch! Thanks for your contribution.

PR #3708, opened 2024-03-25
feat: Add a display bytes option to top-languages card

### LINKED ISSUE #3707: Add an option to display byte count for Top-languages
### Is your feature request related to a problem? Please describe.

Currently, the Top Language Card only displays the percentage of each language used. However, adding an option to display byte counts could incentivize more coding activity, thereby increasing the amount of code generated by individuals.

### Describe the solution you'd like

I have already implemented this on my self-hosted github-readme-stats.
( So, if there is demand for this feature from the community, I can clean up the implementation and submit a pull request :)

This can be achieved by making minor modifications to the `api/top-lang.js` and `src/cards/top-languages-card.js`.

### Describe alternatives you've considered

_No response_

### Additional context

It might require CSS adjustments for many layouts. For instance, the default layout appears as follows:

abap34:
## What's Change ?

Add the `&display_bytes=` parameter to the top-languages card.

(Fix #3707)

## Implementation

1. Add `display_bytes` parameter to `api/top-langs.js` and `src/cards/types.d.ts`.
2. Pass this parameter to each `render~Layout` function. 
3. Implement `formatBytes` in `src/common/utils.js`. This function formats the bytes into a human readable string
4. Write the following code so that the appropriate values are displayed where the progress is actually generated.

```javascript
const percentage = ((lang.size / totalSize) * 100).toFixed(2) + "%";
const bytes = formatBytes(lang.size);

const showValue = displayBytes ? bytes : percentage;
```

## Result

A preview of this PR can be seen at `https://github-readme-stats-git-featuredisplaybytes-abap34.vercel.app/api/top-langs`.

Example:


qwerty541 (changes requested):
Hey, @abap34! Thanks for opening this pull request.

I'd like to make some changes in user API. Let's replace `display_bytes: boolean` query string parameter with `stats_format: "percentages" | "bytes"`.

I think that this way the code will be more scalable. In case if we are going to implement more stats formats for this card in future, like displaying count of repositories where the particular languages was used or count of lines written on particular languages or anything else, then we will be able to extend an existing query string parameter with new enum values instead of bloating the total number of query string parameters with `display_repos_count`, `display_lines_count` etc.

abap34:
Thanks for the review, @qwerty541!

It would be better to implement it as you recommended. I will make the modifications.

By the way, if we implement it this way, I think a display style like "670,404 bytes" can be added. (add options like `bytes_short`, `bytes`)

Displaying the byte count as is is my favorite except that it might break the layout. I think it would be nice to have this display option if there is also a shortened version, so that users can choose a shortened one to preserve the layout.

But of course, stability of the layout is also important.
What do you think about this idea?

qwerty541:
In my opinion it will be better to separate these two features. I'm okay with addition of `bytes` stats format for now and `bytes_long` in future. I always try to adhere the following principle:

Less code changes/features per pull request. => More easier and faster review. => More chances to get another collaborators approval and merge.

Let's keep this pull request more simple since I expect that layout changes will requre quite more code changes.

abap34:
Understood, I comprehend your point.

I will implement `bytes` and `percentages`.

abap34:
I have done the implementation.
(`https://github-readme-stats-git-featuredisplaybytes-abap34s-projects.vercel.app/api/top-langs/?username=anuraghazra&stats_format=bytes` to preview!)



The only change I made was to replace all `displayBytes` with `statsFormat` and error-checking.

src/cards/top-languages-card.js
```diff
@@ -203,20 +204,34 @@ const trimTopLanguages = (topLangs, langs_count, hide) => {
...
+  index,
+}) => {
   const staggerDelay = (index + 3) * 150;
   const paddingRight = 95;
   const progressTextX = width - paddingRight + 10;
   const progressWidth = width - paddingRight;
 
+  const progress = parseFloat(((size / totalSize) * 100).toFixed(2));
+  const bytes = formatBytes(size);
+  const showValue = statsFormat === "bytes" ? bytes : `${progress}%`;
```
qwerty541, line 229: I see that this file contains the almost same code on 266-269 lines. I think it will be better to avoid duplication of this logic and create separate function.

tests/utils.test.js
```diff
@@ -7,6 +7,7 @@ import {
   parseBoolean,
```
qwerty541, line 7: It will be awesome if you can add test of this feature in [renderStatsCard.test.js](https://github.com/anuraghazra/github-readme-stats/blob/master/tests/renderStatsCard.test.js)

readme.md
```diff
@@ -108,6 +108,7 @@ Please visit [this link](https://give.do/fundraisers/stand-beside-the-victims-of
     - [Donut Vertical Chart Language Card Layout](#donut-vertical-chart-language-card-layout)
```
qwerty541, line 108: You also need to add new query string parameter information in table on 404 line

tests/utils.test.js
```diff
@@ -134,6 +135,20 @@ describe("Test utils.js", () => {
       borderColor: "#fff",
     });
   });
+
+  it("formatBytes: should return expected values", () => {
+    expect(formatBytes(0)).toBe("0.0 B");
```
qwerty541, line 140: I have some doubts about that bytes have decimal part. I think it should be just 0 B.

qwerty541 (changes requested):
Hey, @abap34! Thanks for your efforts and patience. I just left a few comments about minor things that needs to be fixed/improved. After these changes I will be ready to give my approval. Special thanks for deployed preview link, this helped to avoid the need to checkout this branch and test locally.

src/cards/top-languages-card.js
abap34: I'd prefer it that way too, modified at https://github.com/anuraghazra/github-readme-stats/pull/3708/commits/0c90270143bdb0aaeab32d9887cbed8b4d0953b7.

readme.md
abap34: Sorry, I did not check enough.
Add it at https://github.com/anuraghazra/github-readme-stats/pull/3708/commits/27da644f14eb1a0cd8daa6557c14e54bc4637a64.

tests/utils.test.js
abap34: Okay. I removed unnecessary decimal parts at 7e17b50

qwerty541 (approved):

qwerty541:
Hello @abap34 

Thank you for your awesome contribution!

PR #3754, opened 2024-05-02
correction: use <span> to correctly align the cards side by side

muroko:


qwerty541 (commented):
Hey @muroko.

Choice between `a` and `span` tags does not affect on the alignment. As I see, it works with both. Therefore, I do not see any neccesity in these changes.

Also note that `href` attribute doesn't work with `span` tag,

Thats the reasons why I'm closing this pull request. Feel free to reopen and add more clearness about why you think the changes is neccesary.

PR #3757, opened 2024-05-05
made Nepali translation more readable

rishicares:
The previous words used in the Nepali translation were not readable/understandable. I've tried to make it understandable and more sensible.


rickstaa:

PR #3835, opened 2024-07-02
Added dark-smoky style

MaksimSazanovich:


Jay-Karia (changes requested):
Remove '#'s from color declaration strings.

It should look like this:
`title_color: "edf5e1"`
and so on...

PR #3865, opened 2024-07-25
refactor: remove check for private user repository

jonnrib:
This commit removes the check that throws an error if the user's repository is not found or is private. This change is intended to improve the system's handling of user repositories by no longer failing when encountering private repositories.


jonnrib:
Aprovado.

PR #3890, opened 2024-08-07
fix: typo in `themes/README.md`

Xerxes-2:
`&theme=` is mistyped as `?theme=`


qwerty541 (commented):
Hello @Xerxes-2 

Thanks for opening pull request! I appreciate your interest in our project.

The symbol can be different depending on the position in query string, so I don't think that it matters in this place. I think that we can keep it as is.

PR #3939, opened 2024-09-20
feat: whitelist

kyleaupton:
This PR adds a whitelist concept so the api will only render things you want it to.

### Motivation
I am self-hosting an instance of this project on a VPS that doesn't have much resources. And because of that I would like to ensure that only cards for my username are rendered. Also, I am using my own PAT and would like to limit the number of times it's used against GitHub's api.

This PR adds two new env vars: `WHITELIST` and `GIST_WHITELIST` that are comma-separated lists. If someone hits a username based endpoint (`/`, `/pin`, `/top-langs`, and `/wakatime`), and `process.env.WHITELIST` has a populated value, then the whitelist will be enforced. Same thing with `/gist` except the whitelist is gist IDs.

P.S.- I noticed that `/wakatime` doesn't enforce the blacklist. Not sure if this is intended or not, just wanted to bring it up.


qwerty541 (approved):

qwerty541:
Hello @kyleaupton 

PR #4142, opened 2025-02-13
fix: not deployable on free plan with current vercel configuration

Lulalaby:
When using the deploy button or forking and importing, as mentioned [here](https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#on-vercel), vercel rejects the deployment, because the limit for the free plan seems to be 1000 (or 1024, dunno the count here).


Venage5603:
it can actually be bumped to 1024 if you feel like it because that it the actual max size

vercel.json
```diff
@@ -1,7 +1,7 @@
 {
   "functions": {
     "api/*.js": {
-      "memory": 1200,
+      "memory": 1000,
```
Lulalaby, line 4:
```suggestion
      "memory": 1024,
```

Lulalaby:
Done!

Issue #4241, opened 2025-05-03
Add Docker configurations for a vercel alternative for self-hosting


Colin-de-Seroux:
### Is your feature request related to a problem? Please describe.

We don't have docker. For small VPS it's difficult to manage space. It's better to have only docker and manage all projects with it. (it's my personal problem)

### Describe the solution you'd like

Add Dockerfile and docker-compose.yml like https://github.com/anuraghazra/github-readme-stats/pull/3775, https://github.com/anuraghazra/github-readme-stats/pull/3577, https://github.com/anuraghazra/github-readme-stats/pull/2447 PR.

Colin-de-Seroux:
PR https://github.com/anuraghazra/github-readme-stats/pull/4242

Phenix333:
PR https://github.com/anuraghazra/github-readme-stats/pull/4289.

maxkratz:
Is there any news on merging this?

The public deployment on Vercel broke the other day, probably because of https://github.com/anuraghazra/github-readme-stats/issues/3851#issuecomment-2237316047.

Phenix333:
> Is there any news on merging this?
> 
> The public deployment on Vercel broke the other day, probably because of [#3851 (comment)](https://github.com/anuraghazra/github-readme-stats/issues/3851#issuecomment-2237316047).

No news. Now that it's in the top features, an admin may take over.

PR #4242, opened 2025-05-03
#4241 Add Docker configurations

Colin-de-Seroux:
It's of https://github.com/anuraghazra/github-readme-stats/issues/4241.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -823,6 +823,21 @@ Since the GitHub API only allows 5k requests per hour, my `https://github-readme
 5.  You're done 🎉
     </details>
 
+## On docker platform
+
+<details>
+  <summary><b>:hammer_and_wrench: Step-by-step guide for deploying on docker</b></summary>
+
+  1.  Fork or clone this repo as per your needs.
+  2.  Add `express` to the dependencies section of `package.json`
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -811,6 +823,21 @@ Since the GitHub API only allows 5k requests per hour, my `https://github-readme
 5.  You're done 🎉
     </details>
 
+## On docker platform
+
+<details>
+  <summary><b>:hammer_and_wrench: Step-by-step guide for deploying on docker</b></summary>
+
+  1.  Fork or clone this repo as per your needs.
+  2.  Add `express` to the dependencies section of `package.json`
```
JPFrancoia, line 832: You can probably do this in the Dockerfile:

```docker
RUN npm install express.js
```

That way you don't add an extra dependency to the project when this dependency is necessary for the docker deployment only

JPFrancoia: You can probably add this to the Dockerfile:

```docker
RUN npm install express
```

That way you're not adding an extra dependency to the package.json file, if this dependency is only required for the Docker deployment

Phenix333: Hello, that would be more optimized. I foolishly copied the way it was done for Vercel. I'll take care of it this evening and create a new PR.

PR #4343, opened 2025-07-23
fix: handle apostrophes for names ending in x

### LINKED ISSUE #4232: Card titles do not have the correct possessive form for names ending in "x"
### Describe the bug
In the GitHub Stats card, users with names ending in "x" do not have the correct possessive form of their name in their titles.

For example, a user with the name `Max` is displayed as `Max' GitHub Stats` instead of `Max's GitHub Stats`.

### Expected behavior
Users with names ending in "x" should have a title with apostrophe-s, not just an apostrophe.

Image

[Link to Demo](https://github-readme-stats.vercel.app/api?username=TenType
)
### Additional context
Here are the pertinent lines of code that cause this issue:
https://github.com/anuraghazra/github-readme-stats/blob/25a9dabf13dd8891a96be2421b2b0ed98761f504/src/cards/stats-card.js#L257C1-L259C11

Why was "x" included in this case? It is common to omit the "s" for names ending in "s", but I don't see how this makes grammatical sense for names ending in "x".

Sorry if this is pedantic, but this caught the corner of my eye when I was using this project.

malgokiel:
closes #4232 

This fixes the apostrophe placement in names ending with "x" so it matches the grammar rules.

Test case was also updated.

### DIFF
--- a/src/cards/stats.js
+++ b/src/cards/stats.js
@@ -254,9 +254,7 @@ const renderStatsCard = (stats, options = {}) => {
       theme,
     });
 
-  const apostrophe = ["s"].includes(name.slice(-1).toLocaleLowerCase())
-    ? ""
-    : "s";
+  const apostrophe = /s$/i.test(name.trim()) ? "" : "s";
   const i18n = new I18n({
     locale,
     translations: statCardLocales({ name, apostrophe }),


qwerty541:
Hello @malgokiel 

PR #4345, opened 2025-07-24
Update express.js to support containerized deployments + replicate the behaviour of the vercel app

EstoesMoises:
This will help with containerized deployments and it replicates the behaviour of the vercel app. The vercel app gives you all the github stats under the /api route, this update to the express server replicates the same thing.

- Routes would be accessible through /api
- Port explicitely listens on 0.0.0.0 and not localhost

I've tested this on Coolify and works like a charm, have a look at my stats for example:
https://githubstats.estoesmoises.me/api?username=estoesmoises

### DIFF
--- a/express.js
+++ b/express.js
@@ -7,10 +7,17 @@ import gistCard from "./api/gist.js";
 import express from "express";
 
 const app = express();
-app.listen(process.env.port || 9000);
+const router = express.Router();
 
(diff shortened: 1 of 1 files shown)


qwerty541:
Hello @EstoesMoises 

Thank you for your contribution

PR #4351, opened 2025-08-10
fix: better top-lang statistics algorithm

Sakurapainting:
- Each repository now contributes equal weight to language statistics
- Prevents large repositories from dominating the results
- Provides more accurate representation of user's language distribution
- Fixes issue where single large repo could skew statistics (e.g., If I have a repository that is almost entirely written in C, the previous top-lang algorithm would increase my C language proportion to even 98%, even though many of my other repositories are not written in C.)
- Avoided the unreasonable phenomenon where "a single large repository determines the overall language distribution."


Sakurapainting:
i also wonder.

filipopo:
So, what's the blocker

Sakurapainting:
Looks like the repo maintainers don't check PRs

PR #4369, opened 2025-08-22
docs: remove outdated translations

qwerty541:


rickstaa:
Makes sense to me since these are very outdated. We could switch to AI translations by https://store.crowdin.com/translate-readme.

qwerty541:
> Makes sense to me since these are very outdated. We could switch to AI translations by https://store.crowdin.com/translate-readme.

Yeah. I remeber that we considered crowdin integration a long time ago, but I think that there is two problems with it.

The first one is that having documentation translation in repository feels redundant, taking into account that Google Chrome and all modern browsers includes built-in translations features. Besided there is also a lot of free AI tools, so I can't imagine that translating documenation can be a problem for someone.

The second one is that, as far I remember, the integration of crowdin requires some actions on Anurag's side, since there is no connection with him, most likely we will not be able to implement that.

However, I'm open for considering crowdin translation in our new repository.

PR #4371, opened 2025-08-22
FEAT: Added new height and width query parameters

seekayel:
# Purpose
When using github-readme-stats generated assets in situations that need specific aspect ratios (eg social preview for websites) it is helpful to lock the generated image to a specific size.

This PR adds new query parameters of `height` and `width` which lock the produced svg size to all stats card types.

Example: To match social card 1200x630 aspect ratio:
https://github-readme-stats.vercel.app/api?username=anuraghazra&width=600&height=315

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "github-readme-stats",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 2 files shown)


seekayel:
I'm willing to keep the PR updated with base and make any requested changes if it is valuable to the project. However not sure I want to do that if its not of interest.

Let me know if this has a shot of getting merged.

PR #4385, opened 2025-08-26
docs: make table of contents opt-in

qwerty541:


rickstaa (approved):
LGTM but i do think we can simply remove it all togheter since GitHub has its own TOC.

PR #4435, opened 2025-09-18
Fix wakatime

### LINKED ISSUE #4432: Compact layout grid order between WakaTime and Top Languages cards is inconsistent
### Describe the bug

When using the compact layout for both WakaTime stats card and Top Languages card, the grid display is inconsistent:

- Both cards display as N rows and 2 columns in compact layout.
- WakaTime card fills the grid horizontally (row-first): 1-1 → 1-2 → 2-1 → 2-2 → ...
- Top Languages card fills the grid vertically (column-first): 1-1 → 2-1 → 3-1 → 1-2 → 2-2 → ...

This difference results in the two cards displaying the same data in visually different orders, which can be confusing for users expecting consistent sorting and layout behavior.

### Expected behavior

_No response_

Image

### Additional context

_No response_

bbbugg:
This pull request updates the layout logic for rendering the compact Wakatime card, focusing on improving the spacing and alignment of language nodes. The main changes introduce a configurable `lineHeight` parameter, ensuring better vertical spacing and more maintainable code when displaying language statistics.

**Compact layout improvements:**

* Added a `lineHeight` parameter to the `createLanguageTextNode` function, allowing dynamic adjustment of vertical spacing between language nodes.
* Updated the calculation of the card's height in the compact layout to use the new `lineHeight` variable for more accurate sizing.
* Passed the `lineHeight` value from the `renderWakatimeCard` function into `createLanguageTextNode`, ensuring consistent spacing throughout the card rendering process.
* Keep the compact layout grid order same between WakaTime and Top Languages cards.
* fixed #4432

### DIFF
--- a/src/cards/wakatime.js
+++ b/src/cards/wakatime.js
@@ -87,21 +87,24 @@ const createCompactLangNode = ({ lang, x, y, display_format }) => {
  * @returns {string[]} The language text node items.
  */
 const createLanguageTextNode = ({ langs, y, display_format }) => {
+  const halfLength = Math.ceil(langs.length / 2);
+
   return langs.map((lang, index) => {
-    if (index % 2 === 0) {
+    if (index < halfLength) {
       return createCompactLangNode({
         lang,
         x: 25,
         y: 12.5 * index + y,
         display_format,
       });
+    } else {
+      return createCompactLangNode({
+        lang,
+        x: 230,
+        y: 12.5 * (index - halfLength) + y,
+        display_format,
+      });
     }
-    return createCompactLangNode({
-      lang,
-      x: 230,
-      y: 12.5 + 12.5 * index,
-      display_format,
-    });
   });
 };
 
--- a/src/cards/wakatime.js
+++ b/src/cards/wakatime.js
@@ -84,9 +84,10 @@ const createCompactLangNode = ({ lang, x, y, display_format }) => {
  * @param {WakaTimeLang[]} args.langs The language objects.
  * @param {number} args.y The y position of the language node.
  * @param {"time" | "percent"} args.display_format The display format of the language node.
+ * @param {number} args.lineHeight The line height for spacing.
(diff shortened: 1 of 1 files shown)


qwerty541 (commented):
Hello @bbbugg 

Thank you for opening pull request. I appreciate you interest in our project.

Regarding the changes, the wakatime card works as is for years and thousands of users are okay with current behavior. This means that we definitely should not change the default baviour of the card. For preventing users to experience unexpected changes in their profiles withour actions from their side let's keep it as is.

However, I'm open to consider the addition of new query string parameter, for example `&ordering=horizontal | vertical`, to provide more customization ways for users who really wants it.

Also I didn't get the changes related to the line height, I would be grateful in case you can explain in more details.

bbbugg:
> Hello @bbbugg
> 
> Thank you for opening pull request. I appreciate you interest in our project.
> 
> Regarding the changes, the wakatime card works as is for years and thousands of users are okay with current behavior. This means that we definitely should not change the default baviour of the card. For preventing users to experience unexpected changes in their profiles withour actions from their side let's keep it as is.
> 
> However, I'm open to consider the addition of new query string parameter, for example `&ordering=horizontal | vertical`, to provide more customization ways for users who really wants it.
> 
> Also I didn't get the changes related to the line height, I would be grateful in case you can explain in more details.

@qwerty541  Thank you for your review!

About the line height changes:

Previously, both the line spacing and the overall card height were determined by a fixed value of 25, so the line_height parameter had no real effect. In this PR, the spacing between lines and the card height are now both dynamically calculated based on the line_height parameter, making it actually functional.

Additionally, the ordering logic has been refactored to use the `&ordering=horizontal | vertical` parameter.

All changes can be demonstrated through the examples on the top.

bbbugg:
Hi @qwerty541,

Could you please take a look at this PR when you have a chance?
Any feedback would be greatly appreciated.

Thanks

PR #4458, opened 2025-09-23
add card_width parameter to wakatime card

martin-mfg:
Add parameter `card_width` to wakatime-card. This parameter already exists for stats-card and top-languages-card.

#### Default Layout
a1
a2
a3

#### Compact Layout
b1
b2
b3


qwerty541 (approved):

qwerty541:
Hello @martin-mfg 

I have merged this since it generally works as expected, but I would be very grateful if you could update the code by removing magic numbers according to the copilot comments to make the code more readable. I already made some refactoring in places where I was able to understand the numbers purpose, but it's looks not enough.

PR #4468, opened 2025-09-29
Remove `jest-bench` dependency and use custom solutions

harryzcy:
[`jest-bench`](https://github.com/pckhoi/jest-bench) is last updated 2 years and does not support newest `jest` v30. Also, it relies on [`benchmark.js`](https://github.com/bestiejs/benchmark.js) which is not maintained since 2018, and archived in 2024.

To prepare `jest` upgrade, we should first remove `jest-bench`

### DIFF
--- a/tests/bench/calculateRank.bench.js
+++ b/tests/bench/calculateRank.bench.js
@@ -3,7 +3,7 @@ import { it } from "@jest/globals";
 import { runAndLogStats } from "./utils.js";
 
 it("calculateRank", async () => {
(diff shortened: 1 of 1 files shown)


qwerty541 (approved):

qwerty541:
Hello @harryzcy 

Thank you for your contribution.

I can't say that I feel happy about adding more code which would require maintenance. I would definitely prefer to use some external solution. But after quick research I didn't found any benchmarking tool which would work with jest out of the box. Then let's have this as temporary solution.

PR #4469, opened 2025-09-29
Added: Generic Error handling for retryer

Shishir-1995:
Added a generice error handler for retryer.js.


Shishir-1995:
For any uncaught error: It will show the status as UNKNOWN_ERROR with unknown error as message

PR #4494, opened 2025-10-01
docs(readme): fix environment variables link inside common options section

opheliagoldstein:


opheliagoldstein (commented):
:heavy_check_mark:

PR #4500, opened 2025-10-02
fix: make Bulgarian total commits translation shorter to prevent items overlapping

opheliagoldstein:


opheliagoldstein (commented):
:heavy_check_mark:

PR #4513, opened 2025-10-04
chore: remove package lock from vercelignore to prevent discrepancies between dev and prod envs

alexandr-garbuzov:


alexandr-garbuzov (commented):
:heavy_check_mark:

qwerty541 (approved):

PR #4514, opened 2025-10-05
Update short number format

lulunac27a:
Update short number format by adding 2 decimal places instead of 1 for numbers between 1,000 and 9,999


lulunac27a:
Please do that by adding `number_precision` parameter.

lulunac27a:
Better with higher precision value of 2 digits.

qwerty541:
Hello @lulunac27a 

I have added the `&number_precision=2` parameter. You will be able to pass the value from 0 to 2 inclusively to enforce specific decimal precision. The default behavior remained as is.

PR #4532, opened 2025-10-06
refactor: move cache seconds calculation logic into reusable function

alexandr-garbuzov:


alexandr-garbuzov (commented):
:heavy_check_mark:

qwerty541 (approved):

PR #4580, opened 2025-10-15
Updated Readme.md

Sudhss:
made the readme header slightly better, doesn't make a huge difference but i think it looks great


qwerty541 (approved):

qwerty541:
Hello @Sudhanshu-shukl 

I can agree only with that the bigger main title looks better.

Thank you for your contribution.

PR #4646, opened 2025-11-12
fix: prevent TypeError crashes on network errors using safe property access

duvvuvenkataramana:


duvvuvenkataramana:
@anuraghazra review and merge the PR

Issue #4671, opened 2025-12-03
Replace broken images in README


eonist:
Several images in the README are currently broken and no longer render correctly.

This appears to be related to the [public instance](https://github.com/anuraghazra/github-readme-stats/issues/3851) being shut down after the Vercel sponsorship ended. As a result, image URLs that depended on that instance are no longer valid.

**Proposed solution:**
* Update the README to use the new [GitHub Action–based](https://github.com/marketplace/actions/github-readme-stats-action) setup where applicable.
* Replace or remove broken image links.
* Ensure all images reference stable, long-term URLs.

**Alternative solution:**
* We could also use the new [whitelist](https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-environment-variables) feature to deploy a limited vercel instance to generate the card.

**Expected outcome:**
The README renders correctly with no broken images and accurately reflects the current recommended usage of the project.

sermuns:
I don't think there's anything special about the README. The hosted Vercel instance is simply down for now. See #4658.

eonist:
Yes. That's very common with the stats card widgets. As such would it not make more sense to have static images / svgs. It's not exactly transient it will happen again, a lot. In my experience.

eonist:
Or like a fallback static image/svg when the backend is down maybe?

jfullstackdev:
I do agree with that, if the recent image is down, at least display the last generated image

ArsenTech:
> Or like a fallback static image/svg when the backend is down maybe?

I agree with that. It makes the project way better

eonist:
Maybe also not only for the readme.md but project wide as well? Like if my widget goes down, like every other day 😅. Then it could fallback to some hosted image somewhere, of the latest state. like some GitHub cache or smth. I dunno how hard this would be or if it's even feasible, but it would stabilise things. Currently I took down most of the widgets that kept failing. Going back to smaller widgets like shield.io badges and my own custom GitHub Action based hosted widgets.

rickstaa:
@eonist thanks for your issue. I updated the description to better reflect what we want to do. If someone wants to take this on please let me know and I will assign you,

rickstaa:
[...]

PR #4716, opened 2025-12-27
docs: add github action as recommended usage method

rickstaa:
Add new github action as recommended usage method now that we are no longer sponsored by vercel and the hosted endpoint is not available anymore.


rickstaa:
@qwerty541 let's put this in draft. As pointed out by @jfullstackdev in [this comment](https://github.com/anuraghazra/github-readme-stats/issues/4671#issuecomment-3694043970) we should ensure people know that both the action and the self deploymnet are now the recommended methods. I will update my pr.

anuraghazra (approved):

PR #4732, opened 2026-01-03
fix: deprecation warning and test precision

bbrewington:
## Summary

This PR addresses two minor issues:

1. Fix `url.parse()` deprecation warning

    Node.js shows a deprecation warning when running the project:
    `[DEP0169] DeprecationWarning: url.parse() behavior is not standardized and prone to errors...`

    This originates from the proxy-from-env dependency. Since we can't modify the dependency directly, this PR adds a patch-package patch that replaces the deprecated `url.parse()` with the WHATWG URL API (`new URL()`).

    **Relevant files changed**:
    - package-lock.json
    - package.json
    - patches/proxy-from-env+1.1.0.patch

2. Fix floating-point test precision in calculateRank.test.js

    The "beginner user gets B- rank" test fails with a 1 ULP (Unit in Last Place) difference:
    - `Expected: 65.02918514848255`
    - `Actual:   65.02918514848257`

    **Root cause analysis**: The test expectation was computed using a mathematically equivalent but floating-point-different formula than what the actual code produces:

    | Expression                         | Result            |
    |------------------------------------|-------------------|
    | Code: (1 - sum/12) * 100           | 65.02918514848257 |
    | Test expectation: 100 - 100*sum/12 | 65.02918514848255 |

    These are algebraically identical but produce different IEEE 754 results due to different intermediate rounding. Both values are correct to ~14 significant digits. The fix updates the expected value to match what the code actually computes.

    **Relevant files changed**:
    - tests/calculateRank.test.js

## Test Results
  - [x] All 242 tests pass
  - [x] ESLint passes
  - [x] Prettier check passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "github-readme-stats",
       "version": "1.0.0",
+      "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "axios": "^1.13.1",
@@ -38,6 +39,7 @@
         "lint-staged": "^16.2.7",
         "lodash.snakecase": "^4.1.1",
         "parse-diff": "^0.11.1",
+        "patch-package": "^8.0.1",
         "prettier": "^3.7.3"
       },
       "engines": {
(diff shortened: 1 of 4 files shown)


bbrewington:
@anuraghazra just pinging as reminder of this PR.  This was also called out in the PR 4863 above (not using "#" because it's already related)

PR #4743, opened 2026-01-14
Add vercel_deployment_instruction.yaml file

Stewie-pixel:
This yaml file is for developers who can not display the github stats in their README file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .vercel
-.env
 node_modules
 *.lock
 .idea/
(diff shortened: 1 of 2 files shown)


diogocouto18 (commented):
This doesn't actually add self-hosting instructions for this project — despite the filename, `vercel_deployment_instruction.yaml` contains Markdown (not valid YAML deployment config) that points users to an unrelated third-party fork (`Stewie-pixel/github-vercel-deployment`) and asks them to enter their personal GitHub token there.

I deployed my own instance on Vercel today directly from this repo (fork + `vercel deploy`, `PAT_1` env var) and it works fine without needing any third-party intermediary. Routing users' personal access tokens through an unrelated fork adds real risk for no benefit — I'd suggest closing this in favor of the existing self-hosting docs in the main readme, which already cover this correctly.

PR #4784, opened 2026-02-08
fix: prevent server crash when GitHub API returns null data

daniwebdevid:
### Description
Fixed a runtime error where the server would crash with `Cannot read properties of null (reading 'slice')`. This happens when the GitHub API returns null data (common during rate limits or internal API errors).

### Changes
- Added optional chaining and fallback empty arrays to `repoNodes` in fetchers.
- Ensured the app handles empty responses gracefully without crashing.

### Verification
- All 242 tests passed locally.
- Verified that the server no longer throws a 500 error when data is missing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3421,18 +3421,6 @@
         "url": "https://opencollective.com/eslint"
       }
     },
-    "node_modules/eslint/node_modules/@eslint/js": {
-      "version": "9.39.1",
-      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-9.39.1.tgz",
-      "integrity": "sha512-S26Stp4zCy88tH94QbBv3XCuzRQiZ9yXofEILmglYTh/Ug/a9/umqvgFtYBAo3Lp0nsI/5/qH1CCrbdK3AP1Tw==",
-      "dev": true,
-      "engines": {
-        "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
-      },
-      "funding": {
-        "url": "https://eslint.org/donate"
-      }
-    },
     "node_modules/eslint/node_modules/escape-string-regexp": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-4.0.0.tgz",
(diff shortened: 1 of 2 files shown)


martin-mfg:
Hi @daniwebdevid, thanks for the contribution!

[retryer.js](https://github.com/anuraghazra/github-readme-stats/blob/master/src/common/retryer.js) already has various ways to detect invalid responses. Can you provide a reproducible example how to get invalid data from the GitHub API which is not yet detected by retryer.js?

It would probably be best to handle additional error cases in retryer.js too. And we should also have a test for this new error handling.

daniwebdevid:
Hi, thanks @martin-mfg  for the response!

In my case, the API didn't return a standard error message. Instead, it seems to return a null value for certain user data, which then triggers the Cannot read properties of null (reading 'slice') error in the frontend/SVG output.

Screenshot_2026-02-08_15-43-13

I've attached a screenshot of the SVG output I received. It shows the error message rendered directly inside the SVG box. I will try to investigate retryer.js to see why this specific null response isn't being caught by the current logic.

martin-mfg:
I noticed you're using the service at `github-contributor-stats.vercel.app`. This repository here does not contain the code for `github-contributor-stats.vercel.app`. Please discuss any issues with `github-contributor-stats.vercel.app` in the corresponding repository https://github.com/HwangTaehyun/github-repository-contribution-stats.

Alternatively you can use [github-readme-stats-action](https://github.com/stats-organization/github-readme-stats-action) or [github-stats-extended](https://github.com/stats-organization/github-stats-extended).

Btw, out of curiosity: Where did you get to know about HwangTaehyun/github-repository-contribution-stats?

daniwebdevid:
Hi @qwerty541, thanks for asking! I actually found this repo while investigating an issue with my README profile, which I generated using GPRM. I noticed the stats weren't rendering correctly and tracked the error back to the API response handling.

I realized the null (reading 'slice') crash was happening because of how the fetchers in this repository process data when the GitHub API is under rate limit or returns empty results. Even though I saw the other service, I believe fixing the logic here is important for the stability of this project. I've tested it locally and it fixes the crash, so I hope it can be merged to help others facing the same issue. Cheers!

PR #4714, opened 2025-12-25
feat: Add Cloudflare Workers support

biplavbarua:
# Cloudflare Workers Support

## Description
This PR enables `github-readme-stats` to be deployed on **Cloudflare Workers**. 

This is a strategic addition to address the "Deployment Paused" issues caused by Vercel's free tier limits. Cloudflare Workers provides 100,000 requests/day on its free tier, offering a robust alternative for the community.

## Changes
- **New Adapter Layer (`src/common/adapter.js`)**: Translates standard Request/Response objects to the Express-style `req`/`res` expected by existing handlers.
- **Worker Entry Point (`src/worker.js`)**: Handles routing and dispatches requests to the card generators using the adapter.
- **Wrangler Config (`wrangler.toml`)**: Configuration for Cloudflare deployment, using the `nodejs_compat` compatibility flag.
- **JSON Imports**: Updated `gist.js` and `wakatime.js` to use standard ES imports for JSON, replacing `createRequire` to support modern bundling.

## Verification
- Verified locally using `wrangler dev` (port 8787).
- Confirmed that routing to `/api` works and invokes the stats fetcher (returning error SVG when tokens are missing, which proves execution path).
- Existing tests passed.

## How to Test
1. `npm install`
2. `npm install -D wrangler` (if not present)
3. `npx wrangler dev`
4. Access `http://localhost:8787/api?username=anuraghazra`

Closes # (if applicable)

PR #4823, opened 2026-02-28
feat: Add Cloudflare Workers support + follow-up fixes

BBleae:
## Summary
- include Cloudflare Workers support from #4714
- fix runtime HTTP behavior to use native `fetch` in non-test environments
- normalize non-GraphQL HTTP errors into GraphQL-like `errors` shape
- lazily load worker routes after env injection
- preserve `NODE_ENV=test` in pat-info tests

## Context
This branch contains commit `1c4a4b5` from #4714 plus follow-up commit `727d211`.

I could not push to the original PR head branch (`biplavbarua:master`) due permissions, so this PR is opened as a replacement path for review.

Related: #4714

### DIFF
--- a/wrangler.toml
+++ b/wrangler.toml
@@ -1,0 +1,8 @@
+name = "github-readme-stats"
+main = "src/worker.js"
+compatibility_date = "2024-09-23"
+compatibility_flags = ["nodejs_compat"]
+
+[vars]
+# Add your PATs here or via `wrangler secret put`
+# PAT_1 = "..."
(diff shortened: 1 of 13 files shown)


wrangler.toml
```diff
@@ -0,0 +1,8 @@
+name = "github-readme-stats"
+main = "src/worker.js"
+compatibility_date = "2024-09-23"
```
harryzcy, line 3: Use a more recent date: `2026-06-16`

harryzcy:
Related issue: https://github.com/anuraghazra/github-readme-stats/issues/3433

PR #4880, opened 2026-05-07
Fix stats card label/value overlap for long labels

tuconnaisyouknow:
## Summary

This fixes a layout issue on the stats card `/api` where long labels such as `Total Commits (last
year):` could overlap the numeric value.

## Root cause

The value column used a fixed horizontal position, which was not robust enough for longer translated
labels or layouts with `show_icons=true` and `hide_rank=true`.

## Changes

- compute the required width from the rendered label/value text widths
- reserve a dedicated right-aligned value column
- grow the card width when the content requires more horizontal space
- add a regression test for long labels on the stats card

## Validation

- `npm run lint -- src/cards/stats.js tests/renderStatsCard.test.js`
- `node --experimental-vm-modules node_modules/jest/bin/jest.js tests/renderStatsCard.test.js tests/
api.test.js`

### DIFF
--- a/src/cards/stats.js
+++ b/src/cards/stats.js
@@ -16,68 +16,30 @@ const RANK_CARD_MIN_WIDTH = 420;
 const RANK_CARD_DEFAULT_WIDTH = 450;
 const RANK_ONLY_CARD_MIN_WIDTH = 290;
 const RANK_ONLY_CARD_DEFAULT_WIDTH = 290;
+const STAT_ROW_X = 25;
+const STAT_LABEL_X = 25;
+const STAT_FONT_SIZE = 14;
+const STAT_LABEL_VALUE_GAP = 28;
+const STAT_LABEL_WIDTH_BUFFER = 20;
+const STAT_VALUE_WIDTH_BUFFER = 12;
+const HIDE_RANK_VALUE_RIGHT_PADDING = 25;
+const RANK_CARD_VALUE_RIGHT_RESERVED_WIDTH = 145;
 
 /**
- * Long locales that need more space for text. Keep sorted alphabetically.
+ * Formats a stats card value the same way it is rendered in the SVG.
  *
- * @type {(keyof typeof wakatimeCardLocales["wakatimecard.title"])[]}
- */
-const LONG_LOCALES = [
-  "az",
-  "bg",
-  "cs",
-  "de",
-  "el",
-  "es",
-  "fil",
-  "fi",
-  "fr",
-  "hu",
-  "id",
-  "ja",
-  "ml",
-  "my",
-  "nl",
-  "pl",
-  "pt-br",
-  "pt-pt",
-  "ru",
-  "sr",
-  "sr-latn",
-  "sw",
-  "ta",
-  "uk-ua",
-  "uz",
-  "zh-tw",
-];
-
-/**
- * Create a stats card text item.
- *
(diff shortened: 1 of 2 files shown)


martin-mfg:
Hi, could you please show an example where text overlaps?

I have tried this, and it looks ok to me:
Martin&#39;s GitHub Stats, Rank_ B

tuconnaisyouknow:
> Hi, could you please show an example where text overlaps?
> 
> I have tried this, and it looks ok to me: Martin&apos;s GitHub Stats, Rank_ B

Hi yes, here is a screenshot with the second line and the last line where text overlaps.
2026-06-05-141413_hyprshot

martin-mfg:
In your GitHub profile it looks like you're using `https://github-stats.fayou.cloud/api?username=tuconnaisyouknow&show_icons=true&theme=rose_pine&hide_border=true&hide_rank=true`. I guess you made some changes to the card design on your instance? If so I can't help with that. Otherwise please provide a sample url with overlapping text from https://github-readme-stats.vercel.app/, https://github-stats-extended.vercel.app/ or https://github-readme-stats-fast.vercel.app/.

tuconnaisyouknow:
> In your GitHub profile it looks like you're using `https://github-stats.fayou.cloud/api?username=tuconnaisyouknow&show_icons=true&theme=rose_pine&hide_border=true&hide_rank=true`. I guess you made some changes to the card design on your instance? If so I can't help with that. Otherwise please provide a sample url with overlapping text from https://github-readme-stats.vercel.app/, https://github-stats-extended.vercel.app/ or https://github-readme-stats-fast.vercel.app/.

In my GitHub profile, I use my own instance of `github-readme-stats`, where I already fixed the bug that this pull request addresses.

The bug used to happen with this exact URL, in case you want to reproduce it:

`https://github-readme-stats.vercel.app/api?username=tuconnaisyouknow&hide_title=true&hide_rank=true&show_icons=true&count_private=true&disable_animations=false&theme=rose_pine&locale=en&hide_border=true`

martin-mfg:
> The bug used to happen with this exact URL, in case you want to reproduce it:

> https://github-readme-stats.vercel.app/api?username=tuconnaisyouknow&hide_title=true&hide_rank=true&show_icons=true&count_private=true&disable_animations=false&theme=rose_pine&locale=en&hide_border=true

For reference, here's the result. I don't see any overlaps in Edge or Firefox:
stats_card_firefox

If you do, do you use any customizations or addons in your browser which might cause this?

tuconnaisyouknow:
> > The bug used to happen with this exact URL, in case you want to reproduce it:
> 
> > https://github-readme-stats.vercel.app/api?username=tuconnaisyouknow&hide_title=true&hide_rank=true&show_icons=true&count_private=true&disable_animations=false&theme=rose_pine&locale=en&hide_border=true
> 
> For reference, here's the result. I don't see any overlaps in Edge or Firefox: stats_card_firefox
> 
> If you do, do you use any customizations or addons in your browser which might cause this?

Yeah, this is probably specific to my environment. I'm running Arch Linux with Hyprland, using Brave and JetBrains Mono Nerd Font, so there may be some rendering differences compared to more common setups.

That said, I don't think this change introduces regressions for cases where it already worked before. From my testing, the fix preserves the existing behavior while also covering cases that previously rendered incorrectly, so it should remain safe even if this turns out to be environment-specific.

PR #4882, opened 2026-05-10
docs: add Coolify/Nixpacks self-hosting guide

saadih:
Add a `nixpacks.toml` and accompanying readme section so users can self-host github-readme-stats on Coolify (or any other Nixpacks-based platform) via the Public Git Repository option.

The `nixpacks.toml` pins Node.js 22, installs all dependencies, and starts the Express server. Because dev dependencies aren't pruned, `express` (which lives in `devDependencies`) is available at runtime, so the existing `package.json` doesn't need to change.

This complements (rather than replaces) the existing **On other platforms** instructions, which remain accurate for users on platforms that prune dev deps and need to move `express` manually.

### Files changed
- `nixpacks.toml` (new), 7 lines
- `readme.md`, adds an `On Coolify via Nixpacks` section plus a TOC entry

### Tested locally
- `npm ci --ignore-scripts` installs cleanly (593 packages, no errors)
- `node express.js` boots on the configured `PORT`
- All 5 API routes (`/api`, `/api/pin`, `/api/top-langs`, `/api/wakatime`, `/api/gist`) respond with HTTP 200 and valid SVG

### DIFF
--- a/nixpacks.toml
+++ b/nixpacks.toml
@@ -1,0 +1,8 @@
+[phases.setup]
+nixPkgs = ["nodejs_22"]
+
+[phases.install]
+cmds = ["npm ci --ignore-scripts"]
+
+[start]
+cmd = "node express.js"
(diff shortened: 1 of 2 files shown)


msndrbzyd-beep (commented):
F

Issue #3851, opened 2024-07-17
Public deployment paused


shivpat09:
### Describe the bug

So the commands that have been set to show the most used languages or commits and stuff seem to not be working. They come up as the name of the image in blue text with the weird image icon next to it.

### Expected behavior

I expected the charts and data to pop up like normal, or as it is shown in the README.

### Screenshots / Live demo link

![image](https://github.com/user-attachments/assets/fb9ab0bd-0a02-4a73-a0e8-72e503a870e3)

justinvforvendetta:
same here.

PLeonLopes:
Same here, probably cooking an update 👀

CoolCat467:
If you hit https://github-readme-stats.vercel.app/, we get a `DEPLOYMENT_PAUSED` response code.

shivenpatel399:
Ok, that makes sense. Sorry if I caused any trouble, I tried this stats thing for the first time yesterday, so I don't know much about it.

afan0918:
same here

kpryu6:
same here..

qwerty541:
Hey guys! Thanks for your reports. Looks like something was happened with Vercel account of Anurag since each request returns the page on screenshot. Unfortunately only Anurag has access to the account, I sent him a message, now waiting for his response. If you want to make your cards work right now, then you can [deploy your own Vercel instance](https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#on-vercel).

![Снимок экрана от 2024-07-18 05-17-20](https://github.com/user-attachments/assets/27bcfab2-e771-4312-be4e-e2d37e398f31)

skson0x6ab:
same here

hootyhoot:
> Hey guys! Thanks for your reports. Looks like something was happened with Vercel account of Anurag since each request returns the page on screenshot. Unfortunately only Anurag has access to the account, I sent him a message, now waiting for his response. If you want to make your cards work right now, then you can [deploy your own Vercel instance](https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#on-vercel).
> 
[...]

PR #4884, opened 2026-05-16
fix: replace deprecated #gh-dark-mode-only fragments with <picture> element in Responsive Card Theme preview

Gilmoursa:
## What

Two locations in the README had broken images caused by the deprecated `#gh-dark-mode-only` / `#gh-light-mode-only` URL fragment syntax, which GitHub no longer renders:

1. The preview at the top of the **Responsive Card Theme** section (had an additional malformed double anchor in the link href, e.g. `#responsive-card-theme#gh-dark-mode-only`)
2. The live example inside the **"Use GitHub's theme context tag"** `<details>` block

## Fix

Replaced both instances with `<picture>` elements using `prefers-color-scheme` media queries — the same approach the README itself documents as the correct modern method in the **"Use GitHub's new media feature"** subsection.

The catbox.moe screenshots in the Vercel setup guide were also audited and confirmed still live (all returning HTTP 200).

## Testing

Verified locally that the `<picture>` syntax matches the working examples already present in the **"Use GitHub's new media feature"** subsection.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -234,8 +234,17 @@ You can use [GitHub's theme context](https://github.blog/changelog/2021-11-24-sp
 <details>
 <summary>:eyes: Show example</summary>
 
-[![Anurag's GitHub stats-Dark](https://github-readme-stats.vercel.app/api?username=anuraghazra\&show_icons=true\&theme=dark#gh-dark-mode-only)](https://github.com/anuraghazra/github-readme-stats#gh-dark-mode-only)
-[![Anurag's GitHub stats-Light](https://github-readme-stats.vercel.app/api?username=anuraghazra\&show_icons=true\&theme=default#gh-light-mode-only)](https://github.com/anuraghazra/github-readme-stats#gh-light-mode-only)
+<picture>
+  <source
+    srcset="https://github-readme-stats.vercel.app/api?username=anuraghazra&show_icons=true&theme=dark"
+    media="(prefers-color-scheme: dark)"
+  />
+  <source
+    srcset="https://github-readme-stats.vercel.app/api?username=anuraghazra&show_icons=true"
+    media="(prefers-color-scheme: light), (prefers-color-scheme: no-preference)"
+  />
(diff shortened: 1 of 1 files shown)


martin-mfg:
Thanks for the PR, but it seems the problem with the mentioned images is not the dark-mode/light-mode fragment. The problem is that the [server is down](https://github.com/anuraghazra/github-readme-stats/issues/3851) and doesn't serve any images. We can see e.g. [here](https://github.com/stats-organization/github-stats-extended/blob/master/docs/advanced_documentation.md#responsive-card-theme) that the dark-mode/light-mode fragments work.

Or did I miss something?

Gilmoursa:
Good catch on the server outage — you're right that the preview images are currently broken because of that, not purely the fragment syntax. That said, GitHub's documentation does explicitly mark `#gh-dark-mode-only` / `#gh-light-mode-only` as deprecated in favor of the `<picture>` element, so the fix is still valid as a proactive cleanup even if the images won't render until the server is back up. Happy to close if the project prefers to hold off until the deployment situation is resolved.

martin-mfg:
Good point. I agree that we should anyway recommend the `<picture>` approach.

I have created [a PR](https://github.com/stats-organization/github-stats-extended/pull/260) in [github-stats-extended](https://github.com/stats-organization/github-stats-extended/). Because maintenance of this project has moved to https://github.com/stats-organization/github-stats-extended (see https://github.com/anuraghazra/github-readme-stats/issues/3851#issuecomment-3738281321 for context).

> GitHub's documentation does explicitly mark `#gh-dark-mode-only` / `#gh-light-mode-only` as deprecated

Could you point me to this documentation please?

Gilmoursa:
Sorry for the overstatement. It's not explicitly deprecated. The sources I was using were third-party blog posts, not GitHub's docs. The <picture> approach is what GitHub recommends as the modern method, but it's not deprecated. 
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/quickstart-for-writing-on-github#adding-an-image-to-suit-your-visitors

PR #4895, opened 2026-06-08
add deprecation notice to readme

martin-mfg:
As discussed with @rickstaa, this PR adds a deprecation notice to the readme.

You can preview the updated readme [here](https://github.com/stats-organization/github-readme-stats/tree/maintenance-note?tab=readme-ov-file#github-readme-stats).

-----

The readme is quite long, so users might overlook if there's just a 3-line note at the top. The readme also contains calls to action like "Report Bug" at the top, which sounds like the project is actively maintained. To avoid confusion and draw attention to the deprecation notice I moved most of the existing documentation into a expandable section.
Inside an expandable section, GitHub's "warning" sections etc. don't work. So I recreated them with an emoji and a border. If you prefer a different way of changing the readme, please let me know.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -30,6 +30,16 @@
   </a>
 </p>
 
+<p align="center">Love the project? Please consider <a href="https://www.paypal.me/anuraghazra">donating</a> to support the original creator!</p>
+
+> [!CAUTION]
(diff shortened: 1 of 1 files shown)


rickstaa (approved):
@martin-mfg Since this repository has not been actively maintained for some time and the project is still widely used, I think it makes sense to merge this README update so users have a clear path forward.

I have previously tried to discuss moving the repository to an organization and giving you and the other maintainers sufficient permissions, but that has not moved forward yet. This change should help users in the meantime while avoiding continued confusion around the current maintenance status.

@anuraghazra, if you would prefer a different approach, please feel free to revert or adjust this change. If you could set up the appropriate permissions for @martin-mfg and the other maintainers so they can continue working from the upstream repository, that would also be a good outcome.

PR #4930, opened 2026-09-21
Add manual offset env vars for stats not visible to the deploy token

joe-sharp:
Add EXTRA_COMMITS/EXTRA_PRS/EXTRA_PRS_MERGED/EXTRA_REVIEWS/EXTRA_ISSUES, following the existing EXCLUDE_REPO pattern. These let a self-hosted instance add contribution counts from orgs its GitHub token can't (or shouldn't) read directly, such as an employer org where granting the public-facing deploy token SSO access to private repos isn't desirable. Each defaults to 0, so behavior is unchanged unless set.


joe-sharp:
Sorry for the noise. Meant for this to go to my fork

